PR #2, opened 2015-08-31
naming and casing consistency with android example rules

c-parsons:


c-parsons:
We should hold on this change for now, as the new directory structure will conflict with a pending readme.

PR #6, opened 2015-10-14
use ios_application for ios-app tutorial

c-parsons:


damienmg:
What the status of this pull request? It looks good to me but it is quite old. Should we still merge it?

c-parsons:
Yes, we should. (I thought this was already merged, didn't realize this was dangilng..)

FWIW, the tutorial (http://bazel.io/docs/tutorial/ios-app.html) already references ios_application, so this should be merged ASAP

PR #11, opened 2016-02-10
Add explanation of how to find iOS SDK version

kchodorow:


c-parsons:
Actually, we're hoping to infer iOS SDK version when it is unspecified, so for folks who don'twant to target a specific SDK version, this should "just work".

kchodorow:
I tried running without an SDK version specified and got an error that it couldn't load SDK 8.4 (xcodebuild lists 9.2 as my SDK version).  Maybe I can say "If you get <error message>, then you can run..."?

c-parsons:
meh, either way this is fine for now, to be honest. We're at least a week away from the automatic ios sdk detection.

PR #13, opened 2016-03-31
Delete all example but the tutorial

damienmg:
All those examples are unmaintained and dates back from the
alpha version of Bazel.


kchodorow:
LGTM

PR #15, opened 2016-08-26
Implement an example maven project

jin:
Add an example Java project that uses Maven for getting dependencies.

https://github.com/bazelbuild/bazel/issues/1410

I'm in the midst of writing a Skylark rule for maven_jar and maven_server,
but while that is in progress, here's the example project that I'm testing with.

### DIFF
--- a/java-maven/BUILD
+++ b/java-maven/BUILD
@@ -1,0 +1,23 @@
+package(default_visibility = ["//visibility:public"])
+
+java_library(
+    name = "java-maven-lib",
+    srcs = glob(["src/main/**/myproject/*.java"]),
+    deps = ["@com_google_guava_guava//jar"],
+)
+
+java_binary(
+    name = "java-maven",
+    main_class = "com.example.myproject.App",
+    runtime_deps = [":java-maven-lib"],
+)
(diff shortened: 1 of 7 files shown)


java-maven/BUILD
```diff
@@ -0,0 +1,23 @@
+package(default_visibility = ["//visibility:public"])
+
+java_library(
+    name = "java-maven-lib",
+    srcs = glob(["src/main/**/myproject/*.java"]),
```
kchodorow: Prefer non-recursive globs (or no globs at all when there's just one file).

jin: done!

kchodorow:
LGTM.

Jenkins, test this please.

damienmg:
We do not have jenkins presubmit for this project. The reason? there is no separate configure step, we should make this project testable with only `bazel test //...` (well would be a good follow up of that change) :p

damienmg:
LGTM too,

this project won't even run on ci.bazel.io because we only test the tutorial.

kchodorow:
Manually built & tested, looks good.

PR #16, opened 2016-09-10
Make the tutorial build more compatible with ci.bazel.io

damienmg:
The way the tutorial is build on ci was a bit off the way we build the other.
These two changes:
  - create a `//:all` target to build that select automatically what to build using select.
  - add a setup script for the android tooling so we can run that as a configure step.


damienmg:
That with another change to ci should make the Tutorial finally green again. I need that change in to test the change to ci.bazel.io

PR #20, opened 2016-11-07
Remove offband ios sdk detection from CI script

c-parsons:
IOS SDK detection is now handled by the xcode_configure repository rule, and thus passing --ios_sdk_version may not match the actual system default.

### DIFF
--- a/tutorial/ci/build.sh
+++ b/tutorial/ci/build.sh
@@ -22,8 +22,6 @@
 #      the android build won't be tested. The 2 other one are defaulted
 #      to, respectively, "22" and "22.0.1".
 
-IOS_SDK_VERSION=$(xcrun --sdk iphonesimulator --show-sdk-version)
-
 # Go to the workspace root
 cd "$(dirname "$(dirname "${BASH_SOURCE[0]}")")"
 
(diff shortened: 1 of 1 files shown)


kchodorow:
Also update README.md and, if necessary, the mention in https://github.com/bazelbuild/bazel/blob/master/site/versions/master/docs/tutorial/ios-app.md.

damienmg:
This script is not used.

c-parsons:
Indeed, based on the actual CI result failing it seems like this script is actually not used. ios-app.md pointed to it, so I assumed this is what we were running on CI. What is the appropriate location?

damienmg:
According to https://github.com/bazelbuild/continuous-integration/blob/master/jenkins/jobs/BUILD#L50

We do not do anything but building //:all for iOS. 

So I would say the whole setup of iOS is incorrect on ci.bazel.io:
https://github.com/bazelbuild/continuous-integration/blob/master/jenkins/github-jobs.sh.tpl#L52

There is actually a TODO :)

(note, ci is taking contribution only on gerrit at https://bazel.googlesource.com/continuous-integration)

PR #21, opened 2017-01-03
Update to use rules_appengine 0.0.5.

katre:
Also updated to use http_archive, which is much more efficient than git_repository.


katre:
This will fix https://github.com/bazelbuild/bazel/issues/2330

katre:
Jenkins, test this please.

steren:
(We will also need to modify the [tutorial content](https://bazel.build/versions/master/docs/tutorial/backend-server.html#add-a-git_repository-rule) to use `http_archive.)

I have 2 problems now: 
I cannot run anymore `bazel-bin/backend/backend --port=12345` is seems that the `--port` parameter is the issue. `bazel-bin/backend/backend` works.

When I run `$WORKSPACE/bazel-bin/backend/backend.deploy test-gae-bazel`, I get:
````
Cannot stat file /private/var/folders/3v/vr09tqmx6t3bkzc046mzdb84007nxl/T/war.BTlcu4Bi/./WEB-INF: No such file or directory
Cannot stat file /private/var/folders/3v/vr09tqmx6t3bkzc046mzdb84007nxl/T/war.BTlcu4Bi/./WEB-INF/lib: No such file or directory
/Users/steren/work/test-bazel-gae/examples/tutorial//bazel-bin/backend/backend.deploy: line 38: /Users/steren/work/test-bazel-gae/examples/tutorial//bazel-bin/backend/backend.runfiles/__main__/../com_google_appengine_java/appengine-java-sdk-1.9.38//bin/appcfg.sh: No such file or directory
````

katre:
Both of these seem to be problems with the appengine_war rule.

The lack of handling flags is part of com.google.appengine.tools.development.DevAppServerMain, which is  itself from appengine-java-sdk-1.9.48/lib/appengine-tools-api.jar.

Building locally, I see that the backend.war has content in WEB-INF/lib (four jar files), so I'm not sure why the deployment script isn't seeing them. It looks (https://github.com/bazelbuild/rules_appengine/blob/master/appengine/appengine_deploy.sh.template) like it's just unzipping the war file, then calling appcfg.sh, so that should be correct unless the structure is wrong.

At this point I'm not sure if we have a real problem with the tutorial, or with the appengine rules themselves, so I will re-assign the issue to someone who knows them better than I do.

kchodorow:
For the dev appserver part, https://github.com/bazelbuild/rules_appengine/pull/41 fixes it.  Looking into the deploy jar issue now.

kchodorow:
For the deploy jar issue, it looks like it's using AppEngine 1.9.38, not 1.9.48 (which is what rule_appengine 0.0.5 specifies).  Is it possible you're using an older version of rules_appengine (maybe overriding the version in your WORKSPACE file)?

katre:
Sent https://bazel-review.googlesource.com/c/8092/ to update the tutorial docs.

PR #22, opened 2017-05-10
Update default Android version to 24

kchodorow:
Stopgap (but not the proper fix) for https://github.com/bazelbuild/continuous-integration/issues/74.

### DIFF
--- a/tutorial/ci/setup_android_repositories.sh
+++ b/tutorial/ci/setup_android_repositories.sh
@@ -26,8 +26,6 @@ function setup_android_repositories() {
 android_sdk_repository(
     name = "androidsdk",
     path = "${ANDROID_SDK_PATH}",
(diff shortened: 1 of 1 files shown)


aj-michael:
I think we can just delete this line. android_sdk_repository will use the highest available API level if none is installed. Same for build_tools_version. I think we should just remove the lines that set api_level and build_tools_version.

aj-michael:
Augh, I think that https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/android/android_integration_test.sh#L389 will only pass with API level 24.

kchodorow:
This isn't used for android_integration_test.sh (different repo!). Removed the lines.

aj-michael:
Ah, you're right! LGTM.

PR #25, opened 2017-06-06
Add visibility attribute to ios-app to allow building //:all to work.

katre:
Also ran buildifier on ios-app/BUILD.


katre:
Bazel CI is broken because of this: http://ci.bazel.io/job/Tutorial/BAZEL_VERSION=HEAD,PLATFORM_NAME=darwin-x86_64/958/console

PR #28, opened 2017-07-20
Add example for template expansion

laurentlb:


rules/expand_template/hello.bzl
```diff
@@ -0,0 +1,23 @@
+# Path of the template file to use.
```
dslomov: Label

laurentlb:
Done

PR #33, opened 2017-07-21
Add more examples from the cookbook.

laurentlb:
- How to create an empty rule
- How to access rule attributes
- How to compute implicit dependencies based on the attributes
- How to use attr.output_list
- How to create an executable rule
- How to create a rule with a declared output


rules/attributes/BUILD
```diff
@@ -0,0 +1,14 @@
+load("//attributes:printer.bzl", "printer")
+
+# Run 'bazel build :some_target' to print its informations.
+printer(
```
dslomov: information

PR #45, opened 2017-11-09
Remove usages of deprecated APIs

mmorearty:
Modify the examples to no longer use deprecated APIs. For example, instead of `ctx.action()`, the examples now use `ctx.actions.run()` and `ctx.actions.run_shell()`.


bazel-io:
Can one of the admins verify this patch?

laurentlb:
Thanks a lot!

PR #46, opened 2017-11-23
update rules_apple so that it works with Bazel at HEAD

lberki:
...so that it works after the removal of a few fields from `ctx.fragments.apple` .


damienmg:
LGTM

PR #47, opened 2017-11-23
Add depsets example.

dslomov:


rules/depsets/foo.bzl
```diff
@@ -0,0 +1,50 @@
...
+def _foo_binary_impl(ctx):
+  foocc = ctx.executable._foocc
+  out = ctx.outputs.out
+  trans_srcs = get_transitive_srcs(ctx.files.srcs, ctx.attr.deps)
+  srcs_list = trans_srcs.to_list()
+  cmd_string = (foocc.path + " " + out.path + " " +
+                " ".join([src.path for src in srcs_list]))
+  ctx.actions.run_shell(command=cmd_string,
+                        inputs=srcs_list + [foocc],
+                        outputs=[out])
```
laurentlb: Can you use ctx.action.run instead?

dslomov: done

laurentlb:
LGTM

PR #53, opened 2018-01-24
Update README.md package => target

Passw:
should be target not package


bazel-io:
Can one of the admins verify this patch?

jin:
Thanks!

PR #56, opened 2018-03-08
Revamp examples for outputs and executable rules

brandjon:
custom_outputs/ and default_outputs/ are replaced by predeclared_outputs/.
executable/ is rewritten.

Both of the new examples are gimmicky, but less trivial, so hopefully they
provide some informative context.

References from the website (rules page, cookbook) will be updated after this is merged.


rules/executable/fortune.bzl
```diff
@@ -0,0 +1,49 @@
...
+    attrs = {
+        "srcs": attr.label_list(
+            allow_files = True,
+            doc = "Input haiku files. Each file must have exactly three lines. "
+                + "The last line must be terminated by a newline character."
+        ),
+    },
+    outputs = {
+        "exe": "%{name}-fortune",
+    },
```
laurentlb: Maybe use ctx.actions.declare_file instead of the outputs map?

brandjon: Done.

PR #58, opened 2018-04-28
Add declare_directory example

achew22:
This example creates a new "sum" rule in the "directory_outputs"
directory that takes a file containing a single line, a number, and adds
it to the deps inputs. Additionally there is a rule "generate_sums" that
declares a directory of files that meet this criteria. My expectation
was that it would be possible to iterate over an attribute on the File
object returned from declare_directory, but there doesn't appear to be
any way to enumerate the files that were created and pass those
individually as files into my run_shell action.

Unfortunately this doesn't seem to be the case.

I attempted to write about this in a thread on [Bazel-discus](https://groups.google.com/d/msgid/bazel-discuss/CAAVj6hx7L-mbnE2qfmOuWnUrmAJxO5UJ5a8DKJqd0tgzALsmqA%40mail.gmail.com)
but I figure code will do a decent job of explaining my question.

Additionally reformat with mdformat.


petroseskinder:
@laurentlb Any progress on this review?

laurentlb:
ping @tomlu

tomlu:
It isn't possible to iterate over directory contents, because (almost by definition) bazel can't know what's in the directory until execution time. If you did know, then you wouldn't need to use a directory.

rules/directory_outputs/sum.bzl
```diff
@@ -0,0 +1,64 @@
...
+
+NumberInfo = provider("number")
+
+def _sum_impl(ctx):
+  inputs = []
+  if ctx.attr.number:
+    inputs.append(ctx.file.number)
+  for dep in ctx.attr.deps:
+    # What do I do here to detect that the provided number is not a single
+    # file, but instead a directory of files?
```
tomlu: If you expect a provider from each dep then it is the generate_sums rule that needs to handle that.

rules/directory_outputs/sum.bzl
```diff
@@ -0,0 +1,64 @@
...
+    # How do I look at the list of files that exist in this directory to filter
+    # out the things that are not appropriate for this rule to digest?
+    inputs.append(dep[NumberInfo].number)
+
+  input_paths = [i.path for i in inputs]
+
+  ctx.actions.run_shell(
+      command = "cat {input_paths} | paste -s -d+ - | bc > {out}".format(
```
tomlu: Use a ctx.actions.args() object and have it expand the directory for you, then your tool reads the files and sums them all.

brandjon:
Removing myself as I don't have bandwidth atm.

Globegitter:
@achew22 Any chance to get this PR updated? Would be great to get a directory example merged into master.

alexeagle:
Doing some cleanup of the repo. It's almost four years since the last update here so I think it went stale. Please comment if that's wrong.

PR #64, opened 2018-06-19
refine bazel-build/examples cpp-tutorial README files

pengwa:


cpp-tutorial/stage3/README.md
```diff
@@ -12,7 +13,8 @@ cc_library(
 )
 ```
 
-Then to use the ```cc_library``` above we add an extra dependency in the form of //[package-path]:[target-name].
+To use our ```hello-time``` libary, an extra dependency is added in the form of //[package-path]:[target-name], in this case, it's ```//lib:hello-time```
```
jin: while you are here, do you mind changing //[package-path]:[target-name] to `//path/to/package:target_name`?

pengwa: yes, sure.

pengwa:
done.

jin:
Thank you so much!

PR #65, opened 2018-06-26
Return DefaultInfo from _foo_library_impl()

nex3:
If we don't do this, transitive uses of foo_library three or more
layers deep fail to compile.

See also bazelbuild/rules_sass#33


jin:
thanks!

PR #76, opened 2018-08-31
Add simple Android example from Bazel's source tree.

jin:
- Updated the project to make use of gmaven_rules/gmaven_artifact()
- Updated README.md to reflect latest API/NDK versions, and usage of
  gmaven_rules
- Cleaned up README.md formatting with new headers
- Changed the C++ source to reflect the new project layout, because the current
  WORKSPACE location has changed.

Related: https://github.com/bazelbuild/bazel/issues/6007

After this, we can remove the Android example from the Bazel soruce tree.

### DIFF
--- a/android/simple/README.md
+++ b/android/simple/README.md
@@ -1,0 +1,88 @@
+# Simple Android Example
+
+## Setup Android SDK and NDK
+
+In order to build these examples, add the following two rules to the top-level
+`WORKSPACE` file (two directories above this file):
+
+```python
+android_sdk_repository(
+    name = "androidsdk",
+    path = "<full path to your Android SDK>",
+    api_level = "<api level>",
+)
+
+android_ndk_repository(
+    name = "androidndk",
+    path = "<path to your Android NDK>",
+    api_level = "<api_level>",
+)
+```
+
+For the `android_sdk_repository` rule, the value of `api_level` corresponds to a
+directory in the SDK containing the specific version of `android.jar` to compile
+against. For example, if `path = "/Users/<username>/Library/Android/sdk"` and
+`api_level = 27`, then the directory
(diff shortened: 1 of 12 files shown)


jin:
ping @ahumesky

android/simple/README.md
```diff
@@ -0,0 +1,88 @@
...
+## Setup Android SDK and NDK
+
+In order to build these examples, add the following two rules to the top-level
+`WORKSPACE` file (two directories above this file):
+
+```python
+android_sdk_repository(
+    name = "androidsdk",
+    path = "<full path to your Android SDK>",
+    api_level = "<api level>",
```
ahumesky: let's update the doc to say that path and api_level are now optional, similarly for android_ndk_repository below

android/simple/java/bazel/BUILD
```diff
@@ -0,0 +1,39 @@
+load("@gmaven_rules//:defs.bzl", "gmaven_artifact")
+
+filegroup(
+    name = "srcs",
```
ahumesky: is this filegroup used anywhere?

android/simple/java/bazel/BUILD
```diff
@@ -0,0 +1,39 @@
...
+    resource_files = glob(["res/**"]),
+    deps = [
+        ":jni",
+        ":lib",
+        gmaven_artifact("com.android.support:appcompat-v7:aar:27.1.1"),
+    ],
+)
+
+cc_library(
+    name = "jni",
```
ahumesky: the ndk example shows how to integrate with native deps, do we need that here for the simple example too? probably would be more valuable to show a more interesting interaction between the android_binary and a more complicated android_library (e.g. the library is in a separate package with resources and its own activity and manifest)

jin:
Dropping this PR in favor of using and expanding the tutorial app.

PR #78, opened 2018-11-08
Add barebones app for the tutorial

jin:
This app does not contain any Bazel files.

DO NOT SUBMIT until tutorial is online.


snazhmudinov:
Why do we need 2 manifest files?

ahumesky:
The tutorial has the code organized into an `android_binary` and a separately `android_library`. `android/tutorial/src/main/java/AndroidManifest.xml` is for the `android_binary`, and `android/tutorial/src/main/java/com/example/bazel/AndroidManifest.xml` is for the `android_library`. The Android manifests from dependencies get merged into the top-level manifest.

PR #87, opened 2019-01-30
Create CODEOWNERS

dslomov:
Adding GitHub repo admins and other interested parties as global CODEOWNERS.


dslomov:
> Could you add a case for the android/ directory for @ahumesky and myself please?

Feel free to add it yourself after we land initial CODEOWNERS

PR #91, opened 2019-02-26
Add presubmit configuration.

fweikert:
Add initial presubmit configuration to be used by https://buildkite.com/bazel/bazel-bazel-examples

It's expected that we need to modify the configuration once we get the first test result from Buildkite.

Fixes https://github.com/bazelbuild/continuous-integration/issues/225.


fweikert:
Actually buchgr@ set up a webhook so that this PR should have triggered a run on Buildkite :(

fweikert:
Nvm, I forgot to check a checkbox on Buildkite.

fweikert:
Current plan: Merge this PR, then find team members that can fix the example projects.

laurentlb:
> Current plan: Merge this PR, then find team members that can fix the example projects.

Sounds good.

PR #96, opened 2019-02-26
Fix FCM example on CI

jin:
Use a fake google-services.json for CI to build.

cc @ahumesky


jin:
@fweikert it's passing now.

fweikert:
\o/ thanks so much!

PR #102, opened 2019-05-28
Add a simple text processing example

brandjon:
I thought this might be handy as dirt-simple example of starlark code that's not too abstracted away from a real build step. And it has docstrings. ;)


brandjon:
The goal of this example is not to highlight any particular feature, but to provide the simplest possible usage of a rule that's not too trivial to demonstrate what a starlark rule actually does.

The example you mention does like very similar in construction (though it uses an output map, not sure whether that makes it easier or harder to understand). What I wanted here though was to have a simple story for how a rule transforms input into output. Maybe I should instead extend the shell example with these rules.

c-parsons:
FWIW, I'm fine with either approach. Both examples demonstrate simple rules quite well, IMO, and aren't quite redundant with each other.

laurentlb:
Maybe add it to the `shell_command` directory as a 2nd example?

We can remove the `outputs = {"out": "%{name}.size"}` (we don't recommend it anymore) from the size example, and create the output with `ctx.actions.declare_file`.

brandjon:
FYI this repository was incorrectly configured to enforce branch protections on **all** branches, including pull request branches. This meant I was unable to push updates to this PR because those updates were themselves not approved by code review. I changed the repo to only enforce any kind of protection on `master`.

brandjon:
I'll redo this as changes to the existing shell rule file.

laurentlb:
> I was unable to push updates to this PR

Usually, contributors are expected to fork the repo to do a PR.

brandjon:
True, I took a shortcut. I'm fine with configuring all our repos to disallow branch creation, since master's the only meaningful branch anyway.

brandjon:
Updated to add to the shell example instead. It's no longer as minimalistic as when I started but hopefully it's pretty explanatory for a newcomer. PTAL

brandjon:
That's reasonable, it doesn't really add much does it. Dropped.

PR #107, opened 2019-07-31
Update Java Maven example to use rules_jvm_external

jin:


jin:
CI failures are fixed in https://github.com/bazelbuild/examples/pull/108

PR #111, opened 2019-08-13
copy hlopko@'s platforms examples to this repo

aiuto:
I am just moving the code over in this PR. I still have to figure out where to place the index from https://docs.google.com/document/d/1UZaVcL08wePB41ATZHcxQV4Pu1YfA1RvvWm8FbZHuW8/edit#heading=h.r33iyda45sjm into the site (or a different site?).


hlopko:
LGTM. Care to also migrate Julie's https://github.com/hlopko/bazel_platforms_examples/pull/1?

rockwotj:
Is there work planned to merge this? These examples are super helpful but where very hard to find.

agoulti:
> Is there work planned to merge this? These examples are super helpful but where very hard to find.

See comments above. There's some issues with these examples.

* yolo_library is not examining the platform that got selected, but only prints what the toolchain gave it. So executing_on_cpu etc are the toolchain's "wishful thinking" and may or may not match the platform that's actually selected. This could lead to misunderstandings.

* the use of HOST_CONSTRAINTS is wrong. It seems to think that this would pick the host platform, where in reality it would pick any platform with the same constraints as host.

I don't think there's anyone actively working on this at the moment.

aiuto:
Yes. The problem is that no one quite has the time to keep the samples
running.

Tyler: Now that you have found them, would you be willing to contribute by
coming up with a PR brought up to current Bazel?

PR #116, opened 2019-08-27
Update rules_apple

keith:
Fixes https://github.com/bazelbuild/bazel/issues/9257


meteorcloudy:
Hmm, this is failing with Bazel@HEAD
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1168#037b7210-2382-4625-973f-f3963f3aae31

Can you take a look?
Also /cc @irengrig

jin:
switching it to http_archive fixed it, but it's not clear why git_repository is failing here. Maybe @irengrig will know more? https://github.com/bazelbuild/examples/pull/118

keith:
Thanks for looking into it, that definitely seems strange to me

PR #119, opened 2019-09-19
Point to executable

nicoaragon:
//actions_run:merge will not hit merge.sh.
The proposed change updates the label so that it goes to //actions_run:merge.sh.

### DIFF
--- a/rules/actions_run/BUILD
+++ b/rules/actions_run/BUILD
@@ -8,10 +8,34 @@ concat(
         "body.html",
         "footer.html",
     ],
+    merge_tool = select({
+        "//conditions:default": "//actions_run:merge_on_linux",
+        "on_linux": "//actions_run:merge_on_linux",
+        "on_windows": "//actions_run:merge_on_windows.bat",
+    }),
 )
 
 # This target is used by the shell rule.
 sh_binary(
-    name = "merge",
+    name = "merge_on_linux",
     srcs = ["merge.sh"],
+)
+
+sh_binary(
+    name = "merge_on_windows.bat",
+    srcs = ["merge.bat"],
+)
+
+config_setting(
+    name = "on_linux",
+    constraint_values = [
+        "@platforms//os:linux"
+    ],
+)
+
+config_setting(
+    name = "on_windows",
+    constraint_values = [
+        "@platforms//os:windows"
+    ],
 )
--- a/rules/actions_run/execute.bzl
+++ b/rules/actions_run/execute.bzl
@@ -18,7 +18,7 @@ def _impl(ctx):
         outputs = [ctx.outputs.out],
         arguments = args,
         progress_message = "Merging into %s" % ctx.outputs.out.short_path,
-        executable = ctx.executable._merge_tool,
+        executable = ctx.executable.merge_tool,
     )
 
 concat = rule(
@@ -26,7 +26,7 @@ concat = rule(
     attrs = {
         "chunks": attr.label_list(allow_files = True),
         "out": attr.output(mandatory = True),
-        "_merge_tool": attr.label(
+        "merge_tool": attr.label(
             executable = True,
             cfg = "host",
             allow_files = True,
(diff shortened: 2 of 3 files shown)


nicoaragon:
@googlebot I signed it!

nicoaragon:
@laurentb @brandonjon tests for openJDK on Windows are failing.
I am not sure I understand what triggered the failure.
What can I do to fix it?

nicoaragon:
I just realized the build needs to change when running on windows. The shell script won't work on for the windows command prompt.

nicoaragon:
The build is [failing in this windows environment](https://buildkite.com/bazel/bazel-bazel-examples/builds/154#716cf6b5-7c2c-42e0-8f37-02c6368e0dd1), but it [succeeds on the remaining environments](https://buildkite.com/bazel/bazel-bazel-examples/builds/154) for windows and linux. The [output says](https://buildkite.com/bazel/bazel-bazel-examples/builds/154#716cf6b5-7c2c-42e0-8f37-02c6368e0dd1/234-259):

> The syntax for the command is incorrect.

I tested the script on "Windows Server version 1803 Datacenter Core", and it worked.

What do you recommend to debug the problem?

rules/actions_run/BUILD
```diff
@@ -8,10 +8,34 @@ concat(
...
+        "//conditions:default": "//actions_run:merge_on_linux.sh",
+        "on_linux": "//actions_run:merge_on_linux.sh",
+        "on_windows": "//actions_run:merge_on_windows.bat",
+    }),
 )
 
 # This target is used by the shell rule.
 sh_binary(
-    name = "merge",
+    name = "merge_on_linux.sh",
```
laszlocsomor: No need for the `.sh` suffix in the target name.

You can use this tool on Windows too (`sh_binary` generates a `.exe` on Windows), just not a `.sh` file directly.

nicoaragon: Thank you. I applied the change for ".bat" too and the build failed for windows again. It says ["target name and extension must match source file extension"](https://buildkite.com/bazel/bazel-bazel-examples/builds/155#fa6ad75b-6bb3-4944-bbd7-764c9d0f9b1b/236-258).

Will try again, trimming ".sh" for linux targets and leaving ".bat" for windows.

nicoaragon: After adding ".bat" for windows, the build for windows is failing with ["the syntax for the command is incorrect"](https://buildkite.com/bazel/bazel-bazel-examples/builds/156#073cf467-7870-4090-a7e2-56a87ac40439/238-262).

I am not sure this is due to the ".bat" file because I verified it on a VM with "Windows Server version 1803 Datacenter Core". It worked there.

What do you recommend as next steps for debugging?

laszlocsomor: I debugged the build and found the problem. Here's how:

1. I cloned this PR, then did `cd rules\actions_run` and `bazel build :sh`
2. I got the "the syntax for the command is incorrect" error too, so I added `--verbose_failures` to Bazel
3. That flag didn't show anything new, so I removed the `@echo off` line from `merge.bat` and built again.
4. Now I saw the failing command was `type  actions_run/header.html actions_run/body.html actions_run/footer.html  1>bazel-out/x64_windows-fastbuild/bin/actions_run/page.html`. I know that `type` requires `\` directory separators, so I edited `merge.sh` like so:
    ```diff
    -type %files% > %out%
    +type %files:/=\% > %out:/=\%
    ```
5. This fixed the build.

nicoaragon: Thank you!
I followed your suggestion. I replaced ```/``` with ```\``` separators and removed ```@echo off```.
The build succeeded.

laszlocsomor: Great!

You can put back `@echo off`. I removed it just for debugging.

nicoaragon: Thank you!

When using ```@echo off``` the build fails for me. Removing it gets the build working again.

What if we skip ```@echo off```?

PR #122, opened 2019-10-17
Use proper header guard

AlexanderPaniutin:
Header guard should be PATH_FILE_NAME_H_ as per Google style guide.

Test: run blaze build/run for all targets from stage 1, 2 and 3.
      No new warnings or issues observed.


aiuto:
I'm folding this in to #243.  Changing the header guard, but not changing the include path. 
Keeping it relative makes it easier to move the examples around and for people to download them to try in whatever location they want.

PR #126, opened 2019-10-22
Run `buildifier --lint=fix -r .`

laurentlb:
This fixes future Bazel incompatibilities.
Updated WORKSPACE files manually.

### DIFF
--- a/rules/WORKSPACE
+++ b/rules/WORKSPACE
@@ -1,0 +1,16 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_python",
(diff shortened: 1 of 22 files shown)


rules/WORKSPACE
```diff
@@ -0,0 +1,16 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
c-parsons: This looks like a brand new file, and therefore outside the scope of this PR's description. Was this previously an accidental omission?

laurentlb: The file was already there, but it was empty.
I'm just adding the dependencies we now need.

laurentlb:
Updated a few more workspace files to fix the tests.

PR #127, opened 2019-11-01
Migrate maven_jar usage to rules_jvm_external

jin:
maven_jar is going away. See https://github.com/bazelbuild/bazel/issues/6799


jin:
Ping @ahumesky

PR #129, opened 2019-11-05
Add extensive runfiles examples

c-parsons:


brandjon:
I'd like to review this but not sure if I'll get to it Wednesday. Feel free to merge without my LGTM and I'll send a follow-up PR if I have suggestions.

rules/runfiles/BUILD
```diff
@@ -10,3 +14,47 @@ execute(
...
+)
+tool_user(
+    name = "tool_user",
+    tool = ":tool",
+)
+
+# 4. Tool with runfiles depending on another tool
+# Same as (3), except adds the complexity of a tool which depends on another tool
+# with its own runfiles. This adds complexity, as the runfiles of the "inner tool"
+# is in a differnt location than if the inner tool were run itself.
```
laurentlb: different

c-parsons:
Looks like the new examples aren't compatible with windows :\ will investigate tomorrow, but CCing @laszlocsomor for FYI (if I can't figure it out tomorrow I may need your help  :))

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+        executable = tool_files_to_run,
+        arguments = [str(my_out.path)]
+    )
+
+    return [DefaultInfo(files = depset([my_out]))]
+
+tool_user = rule(
+    implementation = _tool_user_impl,
+    attrs = {
+        "tool": attr.label(mandatory = True),
```
laszlocsomor: should be `executable = True, allow_single_file = True, cfg = "host"`

https://github.com/bazelbuild/bazel-skylib/blob/9aa308e1ef0f0f0543e3d58f3b3c5488c087c8c1/rules/run_binary.bzl#L65-L76

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+    my_out = ctx.actions.declare_file(ctx.attr.name + "_out")
+
+    # Use the tool's `files_to_run` to describe the tool's
+    # required environment.
+    tool_files_to_run = ctx.attr.tool[DefaultInfo].files_to_run
+
+    ctx.actions.run(
+        outputs = [my_out],
+        executable = tool_files_to_run,
+        arguments = [str(my_out.path)]
```
laszlocsomor: `str()` is unnecessary

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+    attrs = {
+        "command": attr.string(),
+        "_data": attr.label(
+            allow_files = True,
+            default = "//runfiles:data.txt"),
+    },
+)
+
+def _tool_user_impl(ctx):
+    my_out = ctx.actions.declare_file(ctx.attr.name + "_out")
```
laszlocsomor: why not declare it as `ctx.output`?

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+
+    # Use the tool's `files_to_run` to describe the tool's
+    # required environment.
+    tool_files_to_run = ctx.attr.tool[DefaultInfo].files_to_run
```
laszlocsomor: No need for this if you add `executable = True` to the attribute.
Then you can pass `ctx.executable.tool` directly to `ctx.actions.run(executable)`.

https://github.com/bazelbuild/bazel-skylib/blob/9aa308e1ef0f0f0543e3d58f3b3c5488c087c8c1/rules/run_binary.bzl#L49

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+
+    ctx.actions.run(
+        outputs = [my_out],
+        executable = tool_files_to_run,
+        arguments = [str(my_out.path)]
+    )
+
+    return [DefaultInfo(files = depset([my_out]))]
+
+tool_user = rule(
```
laszlocsomor: Looks like a simpler version of [run_binary](https://github.com/bazelbuild/bazel-skylib/blob/9aa308e1ef0f0f0543e3d58f3b3c5488c087c8c1/rules/run_binary.bzl). Maybe you can use that rule instead?

rules/runfiles/tool.bzl
```diff
@@ -0,0 +1,86 @@
...
+    #     <workspace_name>/
+    #         runfiles/
+    #        udata.txt
+
+    # Create the output executable file with command as its content.
+    ctx.actions.write(
+        output = ctx.outputs.executable,
+        # Simple example, effectively puts the contents of data.txt into
+        # the output twice (read once via symlink, once via normal file).
+        content = "cat %s %s > $1" % (data_file_path, data_dep_path),
```
laszlocsomor: This is a shell script, so it'll need a shebang line.
But also, you can't run this on Windows; you must wrap it in a sh_binary.

rules/runfiles/tool.bzl
c-parsons: Because I wanted to demonstrate multiple varied approaches, but I guess these declared files aren't implicitly executable without using `executable = True`,

c-parsons: Whoops -- this comment was for `tool_user` and for not for `tool`. I retract my previous comment and instead say:

"I'd like to demonstrate a varied approach for declaring an output of an action."

rules/runfiles/tool.bzl
c-parsons: Fair enough. Done.

c-parsons: To be frank, I wasn't sure if this was going to flawlessly treat runfiles well, too. But it does! Such magic. I'll update our documentation to suggest *this* approach instead of the alternative :)

rules/runfiles/tool.bzl
c-parsons: I think these (simpler) examples have value over `run_binary` in they explain specifically details of runfiles

rules/runfiles/tool.bzl
c-parsons: Sigh. I wanted to demonstrate a rule which generates a fresh rule, and `sh_binary` muddies things a bit. What are your thoughts on disabling this particular example for windows?

c-parsons:
I'm going to go ahead and commit as-is (with these examples disabled on windows CI), since examples are a pretty low-stakes change cycle. Please feel free to add further remarks and I'll try to address them quickly if you'd like to review further.

rules/runfiles/tool.bzl
laszlocsomor: Good point!

rules/runfiles/tool.bzl
laszlocsomor: Good point. But it may confuse rather than demonstrate: a reader might wonder why this approach here and not the other one. Perhaps add a comment in a follow-up PR, to clarify that this could as well be `attr.output`?

rules/runfiles/tool.bzl
laszlocsomor: As you wish. If you want, I can help you fixing it for Windows later.

PR #142, opened 2020-01-08
repro steps for a bug

nkoroste:
bug: https://github.com/bazelbuild/bazel/issues/10543


alexeagle:
upstream issue was closed 😓

Issue #154, opened 2020-05-11
Add examples of Starlark Configurations usage


juliexxia:
https://docs.bazel.build/versions/master/skylark/config.html

gregestren:
Let's do our best to also link these examples from the core doc (where or if they contextually fit in) before calling this bug finished.

PR #158, opened 2020-05-11
Add examples of selectable copts and test wrapper

juliexxia:
#154


laurentlb:
Checks are failing.

gregestren:
From the error:

`Unable to find package for @bazel_skylib//rules:common_settings.bzl: The repository '@bazel_skylib' could not be resolved.` - odd, since you are including it in the `WORKSPACE`.

To be clear, `cc_binary_selectable_opts` is from https://github.com/gregestren/snippets and `cc_test` is your own new example?

juliexxia:
@laurentlb  - is it right that our CI checks that all targets can be built at the top level with standard configuration? I think I'm getting a lot of failures from that but this example wasn't designed for that to be possible, in fact it's designed to fail if you try to do that. I can rework the example to accommodate but wondering if that's something we can turn off since I imagine other examples will have similar needs.

laurentlb:
You can configure the CI here: https://github.com/bazelbuild/examples/blob/master/.bazelci/presubmit.yml e.g. if you want to pass a flag or disable a target.

I tend to prefer when things can be tested with `bazel test ...`, but I let you decide what's the best here.

gregestren:
Can the library add a default condition that causes it to build with some clear "this is a default mode" implication?

gregestren:
And maybe include some comment along the lines of "*you can comment out the `//conditions:default` if you want your library to require that a depending binary sets an appropriate feature*"?

juliexxia:
I think it makes sense to complicate the yml instead of complicating the model. @gregestren does that sound alright to you? I know it's your example to begin with.  I made the non-building target list quite granular so hopefully we won't miss coverage on anything that's actually supposed to build. Thanks for the tip @laurentlb !

gregestren:
Apologies - who's the next respondent on this PR?

juliexxia:
@gregestren it's ready for you to review again (: I yml-disabled the targets that shouldn't be able to build at the top level so I didn't need to add any default behavior and the examples can run (and fail) as you originally wrote them

juliexxia:
Work towards #154

PR #160, opened 2020-05-12
add example for multi-arch binary

juliexxia:
Final step for https://github.com/bazelbuild/bazel/issues/5575

Note - this logic isn't actually released in bazel yet but I think it's fine to document it prematurely.

### DIFF
--- a/rules/starlark_configurations/multi_arch_binary/defs.bzl
+++ b/rules/starlark_configurations/multi_arch_binary/defs.bzl
@@ -28,6 +28,8 @@ def _rule_impl(ctx):
     # dependencies with providers.
     x86_dep = tools['x86-platform']
     armeabi_v7a_dep = tools['armeabi-v7a-platform']
+    print("transitioned deps: ")
+    print(ctx.split_attr.tool)
     return []
 
 foo_binary = rule(
--- a/rules/starlark_configurations/multi_arch_binary/defs.bzl
+++ b/rules/starlark_configurations/multi_arch_binary/defs.bzl
@@ -1,0 +1,45 @@
+def _transition_impl(settings, attr):
+    _ignore = [settings, attr]
+
+    # Return a dict of dicts. The values are the updates to the configuration.
+    # The keys are arbitrary helpful strings that can be used to access the split
+    # targets in the rule context
+    #
+    # Returning a dict of dicts creates a "split transition", which transitions
+    # the dep its attached to to more than one configuration creating multiple
+    # configured targets.
+    return {
+        "x86-platform": {"//command_line_option:cpu": "x86"},
+        "armeabi-v7a-platform": {"//command_line_option:cpu": "armeabi-v7a"}
+	}
+
+
+fat_transition = transition(
+    implementation = _transition_impl,
+    inputs = [],
+    outputs = ["//command_line_option:cpu"]
+)
+
+def _rule_impl(ctx):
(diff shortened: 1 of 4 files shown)


juliexxia:
I guess checks won't pass until the logic is in bazel...

rules/starlark_configurations/multi_arch_binary/defs.bzl
```diff
@@ -0,0 +1,45 @@
...
+    # dependencies with providers.
+    x86_dep = tools['x86-platform']
+    armeabi_v7a_dep = tools['armeabi-v7a-platform']
+    return []
+
+foo_binary = rule(
+  implementation = _rule_impl,
+  attrs = {
+    "tool": attr.label(cfg = fat_transition),
+    '_whitelist_function_transition': attr.label(
```
gregestren, line 36: Let's comment why this is here, and in every other example that uses this attribute. I know users have been confused by it before.

Do we have a good canonical link to reference that explains this?

rules/starlark_configurations/multi_arch_binary/defs.bzl
```diff
@@ -0,0 +1,45 @@
+def _transition_impl(settings, attr):
+    _ignore = [settings, attr]
+
+    # Return a dict of dicts. The values are the updates to the configuration.
+    # The keys are arbitrary helpful strings that can be used to access the split
+    # targets in the rule context
+    #
+    # Returning a dict of dicts creates a "split transition", which transitions
```
gregestren, line 8: Link to the split transition docs section too?

rules/starlark_configurations/multi_arch_binary/defs.bzl
```diff
@@ -0,0 +1,45 @@
...
+    # the dep its attached to to more than one configuration creating multiple
```
gregestren, line 9: typo nit: it's

rules/starlark_configurations/multi_arch_binary/defs.bzl
```diff
@@ -0,0 +1,45 @@
...
+    outputs = ["//command_line_option:cpu"]
+)
+
+def _rule_impl(ctx):
+    # Access the split dependencies via `ctx.split_attr.<split-attr-name>`
+    tools = ctx.split_attr.tool
+    # The values of `x86_dep` and `armeabi-v7a_dep` here are regular
+    # dependencies with providers.
+    x86_dep = tools['x86-platform']
+    armeabi_v7a_dep = tools['armeabi-v7a-platform']
```
gregestren, line 29: I have an idea that makes the example a bit more elaborate, but could really drive the point home by showing explicitly in the output what's going on.

What about having `_impl` below return a provider with a string based on its CPU, like `SimpleProvider(my_info= "My CPU is X")`. Then have `print()` statements in `_rule_impl` that say `"tools['x86-platform'] provider says: <whatever that instance of the provider says"`.

If it's unclear what I'm saying, let me know and I'll share some more specific code snippets to demonstrate.

gregestren:
Oh, and if `bazelbuild/examples` automatically tries to build whatever gets checked in, will we have a problem if it's not released yet?

gregestren:
Check out https://github.com/bazelbuild/bazel/issues/11077. This might actually get released soon.

juliexxia:
I temporarily disabled it with a TODO back to this issue to re-enable when the release happens

juliexxia:
Added cpu printouts!

PR #162, opened 2020-05-14
Add example on "Make" variables

Wyverald:
for https://github.com/bazelbuild/bazel/issues/11322

after this, we can update the BE to remove the long code blocks (or at least shorten them).

### DIFF
--- a/make-variables/README.md
+++ b/make-variables/README.md
@@ -1,0 +1,33 @@
+# Examples demonstrating "Make" variables
+
+For the complete guide, see the page about ["Make"
+variables](https://docs.bazel.build/versions/master/be/make-variables.html) in
+the BUILD Encyclopedia.
+
+The `//testapp:show_app_output` target demonstrates predefined source/output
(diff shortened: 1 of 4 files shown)


Wyverald:
One thing to perhaps note is that the output files of the two `genrule` targets are now found in blaze-out/ instead of blaze-genfiles/. The latter directory does not even exist. I'm not sure why that's the case since according to https://docs.bazel.build/versions/master/be/general.html#genrule_args the `output_to_bindir` param defaults to False, so the output _should_ be in the "genfiles" directory. Either the documentation is wrong or there is a bug.

laurentlb:
By default in Bazel, the genfiles directory doesn't exist. See https://groups.google.com/g/bazel-dev/c/5iq7n1soenA/m/xGM_M1NOCAAJ for context.

In Google however, we haven't flipped the flags yet. So we somehow have to keep the code/documentation for that, and attributes like `output_to_bindir`.

PR #163, opened 2020-05-14
Add an example for accessing features in a rule.

laurentlb:


laurentlb:
There was a plan in 2017 to have just `ctx.attr.features`: https://docs.google.com/document/d/1wSp1ff_7-f3xXVouayo1zdv-RkzZEc31diy7lPYQTko/edit#heading=h.v8eshpikst2

We can still do it if we want. In the mean time, I'd rather document the current behavior.

PR #167, opened 2020-05-15
Add example of build setting and select

juliexxia:
Work towards #154


rules/starlark_configurations/select_on_build_setting/BUILD
```diff
@@ -0,0 +1,31 @@
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load(":defs.bzl", "harvest", "fruit")
+
+string_flag(name = "season", build_setting_default = "spring")
```
gregestren, line 4: Nit: Might be helpful to comment-contextualize how skylib fits in, i.e. the same comments you made in your other PRs explaining skylib is a convenient but not necessary.

juliexxia: Done

PR #170, opened 2020-05-15
An example of reading settings/attributes in the transition

juliexxia:
...including a note about the reading configured attributed in rule transitions nuance. 

Work towards #154


juliexxia:
Done!

PR #172, opened 2020-05-26
Remove unnecessary header in cpp example (stage 2)

aSquare14:
- This PR removes the header file 
```c
#include<string.h> 
```
from a few files.
- These files imported 
```c
#include "main/hello-greet.h"
```
in which the string header is already present. 
- Hence, importing just hello-greet.h will suffice.


aSquare14:
@googlebot I signed it!

PR #177, opened 2020-06-11
Update execute.bzl

ahumesky:
Make `merge_tool` a private attribute (`"_merge_tool"`), like it says it is in the comments.

### DIFF
--- a/rules/actions_run/execute.bzl
+++ b/rules/actions_run/execute.bzl
@@ -18,7 +18,7 @@ def _impl(ctx):
         outputs = [ctx.outputs.out],
         arguments = args,
         progress_message = "Merging into %s" % ctx.outputs.out.short_path,
-        executable = ctx.executable.merge_tool,
+        executable = ctx.executable._merge_tool,
     )
 
 concat = rule(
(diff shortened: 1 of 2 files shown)


c-parsons:
Failures look legitimate, PTAL?

ahumesky:
yeahhh, looks like making it a public attr was on purpose to support different platforms via a select.... i guess the right thing to do is to create a wrapper macro to create a default select...

ahumesky:
restructured things a little, PTAL

rules/actions_run/execute.bzl
```diff
@@ -30,7 +38,19 @@ concat = rule(
             executable = True,
             cfg = "host",
```
katre, line 39: Also, since you're here, please change this to `cfg = "exec"`.

aiuto:
Friendly ping. Trying merge or close.

aiuto:
Friendly ping. Trying merge or close.

aiuto:
Last ping before closing it.

aiuto:
One final, last ping.  If we are not going to work on this, let's close it.

ahumesky:
I don't really recall what I was doing here, doesn't seem super important

PR #186, opened 2020-11-04
Update rules_apple

keith:
This is required for https://github.com/bazelbuild/bazel/pull/12265


keith:
Maybe we should switch this to use the git repo at HEAD though?

oquenchil (approved):
LGTM

PR #200, opened 2021-04-20
Adding an example for setting multiple fields based on one flag

chasain:
I built this example while trying to make rules_k8s a bit cleaner for us, it allows me to define something like 

```
clusters(name = "cluster", clusters = [":minikube"])
cluster(
  name = "minikube",
  image_chroot = "localhost:5000",
  ...
)
```

Then I can call ```bazel run //some/deployment:object.apply --cluster=minikube``` and substitute all the values from my cluster into the object.

### DIFF
--- a/rules/starlark_configurations/multi_field_string_flag/README.md
+++ b/rules/starlark_configurations/multi_field_string_flag/README.md
@@ -1,0 +1,10 @@
+This is an example of encoding multiple fields into a single string flag. It makes use of starlark aspects to lookup several
+fields based on the value of the flag.
+
+This is useful if you have several settings that are interdependent and you don't want the user to need to define them all on command line.
+
+To test it out, cd to this directory and run the following:
+```
+$ bazel build :ice_cream # => "vanilla is white"
+$ bazel build :ice_cream --//starlark_configurations/multi_field_string_flag:flavor_flag=grape # => "sugar-free grape is purple"
+```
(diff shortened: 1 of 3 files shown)


gregestren:
Thanks for this PR, @chasain . I'm just seeing this now and will give it a thorough review soon.

rules/starlark_configurations/multi_field_string_flag/defs.bzl
```diff
@@ -0,0 +1,81 @@
+# example/buildsettings/build_settings.bzl
+FlavorsProvider = provider(fields = ['flavors'])
+
+def _flavors_impl(ctx):
+    return FlavorsProvider(flavors = ctx.attr.flavors)
+
+flavors = rule(
+    implementation = _flavors_impl,
+    attrs = {
+      "flavors": attr.label_list(),
```
gregestren, line 10: For completeness, I think `attr` can restrict allowed deps to only those that provide `FlavorProvider`, to be more explicit that this is only intended to be consumed by `flavor` instances.

rules/starlark_configurations/multi_field_string_flag/README.md
```diff
@@ -0,0 +1,10 @@
+This is an example of encoding multiple fields into a single string flag. It makes use of starlark aspects to lookup several
+fields based on the value of the flag.
+
+This is useful if you have several settings that are interdependent and you don't want the user to need to define them all on command line.
+
+To test it out, cd to this directory and run the following:
+```
+$ bazel build :ice_cream # => "vanilla is white"
```
gregestren, line 8: Separate the output into distinct lines to look more realistic?

gregestren (commented):
This is a really interesting use of Starlark machinery.

How do you evaluate the importance of aspects in the implementation? Could the `ice_cream` rule itself directly filter its deps' providers based on the flag to get similar results?

aiuto:
Friendly ping.

aiuto:
Another ping, a year later.

Issue #106, opened 2019-07-28
Add examples for the external dependencies


libratiger:
There is no examples about the [external dependencies](https://docs.bazel.build/versions/0.28.0/external.html), it would be better for tutorial.

aehlig:
> There is no examples about the [external dependencies](https://docs.bazel.build/versions/0.28.0/external.html), it would be better for tutorial.

The quoted page shows a couple of examples on how external dependencies can be used; could you please clarify what kind of examples you are missing.

aehlig:
Also, if it is about _this_ repository, the [`WORKSPACE` file of the `tutorial` subdirectory](https://github.com/bazelbuild/examples/blob/8da4b9b4719685d89b840543c74ad5e399b3bcfa/tutorial/WORKSPACE) already uses quite a lot of external repositories; so real-world examples are included here as well.

PR #202, opened 2021-04-28
Add an example on how to deal with third party dependencies

limdor:
There are several ways to handle external dependencies in Bazel.
However there are not many examples on how to address it or what
are the best practices.
In this example one approach is presented.
An example with third party dependencies was requested in #106.
This example has been initially developed in
https://github.com/limdor/bazel-examples, in case there is interest
other examples could be ported.


limdor:
Any feedback would be welcome, thanks

gregestren:
Hi @limdor - I'll try to take a look at this soon. Please ping me again within a week if I haven't responded earlier.

meteorcloudy:
Thanks! This is indeed a nice example and doc of how external dependencies work currently in Bazel for many projects. 

But this approach also has many problems (the major one being it provides poor dependency resolution result). We are working on a completely new external dependency system called "Bzlmod". Please check the the [doc](https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit) and some [examples](https://github.com/meteorcloudy/bzlmod-examples/tree/main/examples). Maybe you could also mention this in the README.md file?

/cc @Wyverald

Issue #198, opened 2021-03-07
Skipping '//main:hello-world': error loading package 'main': Extension file not found


wwdok:
I am following the tutorial step of `cpp-tutorial`, but just at the beginning, i met an error : 
```
C:\MachineLearning\bazelbuild\examples\cpp-tutorial\stage1>bazel build ///main:hello-world
Starting local Bazel server and connecting to it...
ERROR: Skipping '///main:hello-world': not a valid absolute pattern (absolute target patterns must start with exactly two slashes): '///main:hello-world'
WARNING: Target pattern parsing failed.
ERROR: not a valid absolute pattern (absolute target patterns must start with exactly two slashes): '///main:hello-world'
INFO: Elapsed time: 5.773s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)

C:\MachineLearning\bazelbuild\examples\cpp-tutorial\stage1>bazel build //main:hello-world
ERROR: Skipping '//main:hello-world': error loading package 'main': Extension file not found. Unable to load package for '@rules_cc//cc:defs.bzl': The repository could not be resolved
WARNING: Target pattern parsing failed.
ERROR: error loading package 'main': Extension file not found. Unable to load package for '@rules_cc//cc:defs.bzl': The repository could not be resolved
INFO: Elapsed time: 0.318s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
    currently loading: main
```
My computer is windows10, my bazel version is : 
```
C:\MachineLearning\bazelbuild\examples\cpp-tutorial\stage1>bazel version
Build label: 0.19.2
Build target: bazel-out/x64_windows-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Mon Nov 19 16:24:41 2018 (1542644681)
Build timestamp: 1542644681
Build timestamp as int: 1542644681
```

wwdok:
One year later, i pick up learning bazel, even though i run `bazel build //main:hello-world`, this issue still exists like before:
![image](https://user-images.githubusercontent.com/43233772/169645265-76dd7349-f48e-4dfc-9b5a-b4a755700010.png).
[...]

PR #203, opened 2021-05-02
Remove tripple slash on Windows

limdor:
AFAIK there is no longer a need to use tripple slash in Windows, indeed you can have problems if you do:
https://github.com/bazelbuild/examples/issues/198
In case that I am wrong then we should change it to explain in which situation tripple slash might be needed.

### DIFF
--- a/cpp-tutorial/stage1/README.md
+++ b/cpp-tutorial/stage1/README.md
@@ -12,13 +12,9 @@ cc_binary(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


gregestren:
@meteorcloudy - can you sanity check this PR?

@limdor - thanks. Apologies for the delay in looking at this!

meteorcloudy (approved):
Thanks! We no longer need that since a long time ago.

PR #216, opened 2021-07-28
Migrate away from using the 'outputs' parameter

metti:
The `outputs` parameter of rule is deprecated. Hence don't use it in
examples where people are supposed to copy'n'paste from.

Documentation: https://docs.bazel.build/versions/main/skylark/rules.html#deprecated-predeclared-outputs

### DIFF
--- a/rules/actions_write/file.bzl
+++ b/rules/actions_write/file.bzl
@@ -5,12 +5,14 @@ large files with a lot of static content, consider using
 `ctx.actions.expand_template` instead.
 """
 
+def file(**kwargs):
+    _file(out = "{name}.txt".format(**kwargs), **kwargs)
+
 def _impl(ctx):
     output = ctx.outputs.out
     ctx.actions.write(output = output, content = ctx.attr.content)
 
-file = rule(
+_file = rule(
     implementation = _impl,
-    attrs = {"content": attr.string()},
-    outputs = {"out": "%{name}.txt"},
+    attrs = {"content": attr.string(), "out": attr.output()},
 )
--- a/rules/implicit_output/hash.bzl
+++ b/rules/implicit_output/hash.bzl
@@ -24,14 +24,21 @@ def _impl(ctx):
     # source of another target), only the sha256 is computed.
     return DefaultInfo(files = depset([ctx.outputs.sha256]))
 
-hashes = rule(
+_hashes = rule(
     implementation = _impl,
     attrs = {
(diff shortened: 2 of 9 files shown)


rules/implicit_output/hash.bzl
```diff
@@ -24,14 +24,21 @@ def _impl(ctx):
...
+_hashes = rule(
     implementation = _impl,
     attrs = {
         "src": attr.label(mandatory = True, allow_single_file = True),
-    },
-    outputs = {
-        "md5": "%{name}.md5",
-        "sha1": "%{name}.sha1",
-        "sha256": "%{name}.sha256",
+        "md5": attr.output(),
```
lberki, line 31: Whether one should do this depends on whether one wants the user to be able to say e.g. `bazel build :foo.md5` if `hashes(name="foo")` exists. If not, it's better to remove the `outputs=` attribute and use `ctx.actions.declare_file()` in order not to expose the exact name of the files the rule creates.

Given that I haven't seen this idiom before, I suppose `ctx.actions.declare_file()` is the approach to be advertised?

metti: The purpose of this particular example is to declare outputs that need to be manually requested to build them. The default output of this starlark rule is only producing the .sha256 file, unless others are explicitly requested.

From the comments:

```
# To get the md5 sum, run:
# `bazel build //implicit_output:build_hash.md5sum`
# If you use `//implicit_output:build_hash`, you will get only the sha256
```

What am I missing?

lberki: What am *I* missing? :)

I didn't realize that the goal of this example is to demonstrate outputs that are only built when specifically requested. Thanks for the change!

lberki (approved):

rules/actions_write/file.bzl
```diff
@@ -5,12 +5,14 @@ large files with a lot of static content, consider using
...
 def _impl(ctx):
     output = ctx.outputs.out
     ctx.actions.write(output = output, content = ctx.attr.content)
 
-file = rule(
+_file = rule(
     implementation = _impl,
-    attrs = {"content": attr.string()},
-    outputs = {"out": "%{name}.txt"},
+    attrs = {"content": attr.string(), "out": attr.output()},
```
rockwotj, line 17: Don't you need to change line L12 and access this via ctx.attr.out?

metti: Since this is declared as an output it can be accessed through ctx.outputs.out

PR #217, opened 2021-08-08
Update rules_apple

keith:
This should fix this bazel HEAD incompatibility https://github.com/bazelbuild/bazel/issues/13811


philwo:
Thank you!

philwo (approved):

PR #226, opened 2021-10-22
Add Go tutorial

colin-valentini:
Adds a Go tutorial.


aiuto:
Hi Colin.  Are you still interested in working on this? If so, can I ask a few things
- place the code under //tutorial/go rather than at the top level
- Add the text that goes alongside the tutorial to a README.md within it. You might start with the text in https://docs.bazel.build/versions/main/tutorial/cpp.html and work from that.

colin-valentini:
Yes — still interested in working on it. Can make those suggested changes, thank you for taking a look here 👍🏼

alexeagle:
There's now a Bazel Go tutorial at https://bazel-contrib.github.io/SIG-rules-authors/go-tutorial.html by @chrislovecnm
so I think this is duplicate.
Sorry your work went stale @colin-valentini

PR #229, opened 2021-11-22
Examples for flags parsing

aranguyen:
First draft for examples of flags parsing


flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
+Bazel flags parsing examples
+========================
+
+This provides examples for the current behavior of flag/options parsing. This tutorial assumes that users have basic knowledge of creating a Bazel [WORKSPACE](https://docs.bazel.build/versions/main/build-ref.html#workspace) and writing a [BUILD](https://docs.bazel.build/versions/main/build-ref.html#BUILD_files) file. Users should also know what [options/flags](https://docs.bazel.build/versions/main/command-line-reference.html#option-syntax) including built-in (non-Starlark) and user-defined(Starlark) options are.
```
gregestren, line 4: I'd add a link to user-defined flags too: https://docs.bazel.build/skylark/config.html#user-defined-build-settings to clarify these distinctions.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
+Bazel flags parsing examples
+========================
+
```
gregestren, line 3: Let's remove the /versions/main from links. So: https://docs.bazel.build/build-ref.html#workspace. This lets Bazel's website trigger which default version it wants to map links to.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+This provides examples for the current behavior of flag/options parsing. This tutorial assumes that users have basic knowledge of creating a Bazel [WORKSPACE](https://docs.bazel.build/versions/main/build-ref.html#workspace) and writing a [BUILD](https://docs.bazel.build/versions/main/build-ref.html#BUILD_files) file. Users should also know what [options/flags](https://docs.bazel.build/versions/main/command-line-reference.html#option-syntax) including built-in (non-Starlark) and user-defined(Starlark) options are.
+
+Instructions
+========================
+
+```
+cd examples/flags-parsing-tutorial
+```
+In this WORKSPACE, we have:
+* <b>bazelrc file</b>: This is the user-defined bazelrc where options/flags can be defined.
```
gregestren, line 13: Are there any good documentation links to bazelrc?

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+* <b>build_defs.bzl</b>: This contains the Starlark rules' implementations.
+* <b>BUILD</b>: This contains rules Bazel uses to build a package.
+
+For each requirement below, run the following example command and examine output. 
+
+❗❗❗<i>Please note that there is an outstanding bug concerning `--config` and Starlark flags which will be pointed out later in this README. Once a fix is released, this doc will be updated.</i>
+
+### A. Options on the command line takes precedence over those in bazelrc. ###
```
gregestren, line 21: Typo nit: "take" (since "options" is plural). I'm being extra pedantic since this is documentation.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### A. Options on the command line takes precedence over those in bazelrc. ### 
+  
+Without `--config`, this is true for both Starlark and non-Starlark options
+```
+# DEBUG should show cmd for option flag
+bazel --bazelrc=./bazelrc build --//:flag=cmd :flag
+```
+### B. The last flag on the command line takes precedence. ###
+⭐ <b>Tips</b>: Use [--announce_rc](https://docs.bazel.build/versions/main/command-line-reference.html#options-common-to-all-commands) to debug options parsing
```
gregestren, line 29: What about https://docs.bazel.build/command-line-reference.html#flag--announce_rc?

(Sadly the actual documentation for it is super-minimal).

Maybe https://docs.bazel.build/user-manual.html#flag--announce_rc would be the better link?

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+# Since -c is an output affecting option, the "winning" option 
+# will be reflected in the output paths (i.e opt). To see this run `ls -l`
+bazel --bazelrc=./bazelrc build --config=foo -c opt --announce_rc
```
gregestren, line 33: I'd try to be more precise on this one: build an actual target then show how to look at its output file, and observe that it has "opt" in it. I'd also try to avoid use of "output affecting option", since I expect most users won't know what that means. So if we can illustrate the effect with minimal conceptual terms to have to understand, I think that's a good balance.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Here the winning option is that defined in the bazelrc file (i.e dbg).
+# Note that -c=dbg is expanded by --config=foo, the last option on the command line
+bazel --bazelrc=./bazelrc build -c opt --config=foo --announce_rc
```
gregestren, line 37: As before, I'd try to build an actual target. With no targets the build generates an error, which might throw users off ("am I doing this the right way?")

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+bazel --bazelrc=./bazelrc build --//:flag=cmd :flag
+```
+### B. The last flag on the command line takes precedence. ###
+⭐ <b>Tips</b>: Use [--announce_rc](https://docs.bazel.build/versions/main/command-line-reference.html#options-common-to-all-commands) to debug options parsing
+```
+# Since -c is an output affecting option, the "winning" option 
+# will be reflected in the output paths (i.e opt). To see this run `ls -l`
+bazel --bazelrc=./bazelrc build --config=foo -c opt --announce_rc
+
+# Here the winning option is that defined in the bazelrc file (i.e dbg).
```
gregestren, line 35: This is definitely subtle, and it threw me off at first. On first reading I thought this contradicted the observation above that command line options take precedence over bazelrc options. I see it's not actually a contradiction, but could we highlight that subtlety more concretely?

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+# Note that -c=dbg is expanded by --config=foo, the last option on the command line
+bazel --bazelrc=./bazelrc build -c opt --config=foo --announce_rc
+
+# DEBUG should show cmd_last as the "winning" value.
+bazel --bazelrc=./bazelrc build --//:flag=cmd --//:flag=cmd_last :flag
+
+# This is also applicable for cascading --config(s)
+# DEBUG should show flob as value for --//:wibble option
+bazel --bazelrc=./bazelrc build --config=foo --config=bar :wibble :wobble :wubble
```
gregestren, line 44: wibble, wobble, wubble? These are fantastic names!

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+For each requirement below, run the following example command and examine output. 
+
+❗❗❗<i>Please note that there is an outstanding bug concerning `--config` and Starlark flags which will be pointed out later in this README. Once a fix is released, this doc will be updated.</i>
+
+### A. Options on the command line takes precedence over those in bazelrc. ### 
+  
+Without `--config`, this is true for both Starlark and non-Starlark options
+```
+# DEBUG should show cmd for option flag
+bazel --bazelrc=./bazelrc build --//:flag=cmd :flag
```
gregestren, line 26: What would you think of demonstrating the expected output directly, i.e.:

```
$ bazel --bazelrc=./bazelrc build --//:flag=cmd :flag
DEBUG: /my/root/examples/flags-parsing-tutorial/build_defs.bzl:6:10: build setting value for label flag: cmd
```

and keep a light `#` comment to highlight which part of that is interesting?

Then users could compare against exact output if they run it.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+cd examples/flags-parsing-tutorial
+```
+In this WORKSPACE, we have:
+* <b>bazelrc file</b>: This is the user-defined bazelrc where options/flags can be defined.
+* <b>build_defs.bzl</b>: This contains the Starlark rules' implementations.
+* <b>BUILD</b>: This contains rules Bazel uses to build a package.
+
+For each requirement below, run the following example command and examine output. 
+
+❗❗❗<i>Please note that there is an outstanding bug concerning `--config` and Starlark flags which will be pointed out later in this README. Once a fix is released, this doc will be updated.</i>
```
gregestren, line 19: Let's link the bug for better visibility and accountability all around.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,61 @@
...
+# DEBUG should show flob for --//:wibble per A, but instead wibble is shown
+bazel --bazelrc=./bazelrc build --config=foo --//:wibble=flob :wibble
+```
+### C. Within bazelrc file, precedence depends on specificity which is defined by inheritance ###
+Commands such as test and release inherit options from build. The inheriting command is said to be more specific and thus takes precedence.
+```
+# -c is resolved to opt, and --//:wibble resolves to flob.
+bazel --bazelrc=./bazelrc test --config=baz :wibble --announce_rc
+```
+⭐ <b>Tips</b>: For readability, have your most common options at the top of bazelrc.
```
gregestren, line 59: Love all the `Tips`!

flags-parsing-tutorial/bazelrc
```diff
@@ -0,0 +1,19 @@
+### Starlark flag that is not associated with a config
+build --//:flag=flag
+
+### Config foo
```
gregestren, line 4: How about `--config=foo` to make it more clear how this gets invoked?

gregestren (commented):
Overall this looks great. Left a bunch of pedantic comments since it's extra important to be accurate and clear with example docs. 

My main other comment: what if we added a section between the current A and B that describes basic use of --config? While trivial, people looking at Bazel for the first time won't understand the association beforehand. And it sets up the basis for the followup examples.

flags-parsing-tutorial/README.md
aranguyen: Right I see where you're coming from. It threw me off at first as well since flags expanded by config are physically in the bazelrc file. But they are ignored both in bazelrc and command line by default unless --config=<name> is specified. And the --config expanded flags' priority is where --config=<name> is specified (RC priority or command line priority). I will be more clear on this and I think your other suggestion with having the section for --config would help clear this up a little bit.

flags-parsing-tutorial/README.md
aranguyen: haah i didn't know what else beyond foo, bar, baz. So I searched and this is the reference https://en.wikipedia.org/wiki/Metasyntactic_variable

flags-parsing-tutorial/README.md
aranguyen: Ok, let me iterate on this. For the sake of clarity, maybe each example deserves their own bullet point instead of being cramped into one code block as they are right now.

flags-parsing-tutorial/README.md
aranguyen: +1. Thank you for pointing it out!

flags-parsing-tutorial/README.md
aranguyen: Ok, I was focusing more in seeing the options getting resolved. But I agree, users would see a `WARNING: Usage: bazel build <options> <targets>` if the above is run. This would probably confuse users more, I'll add a target for completeness.

flags-parsing-tutorial/README.md
aranguyen: Yup, I can include this https://docs.bazel.build/guide.html#bazelrc-the-bazel-configuration-file

gregestren:
Semi-related, one of my goals for the examples repo is to better standardize templates: what's in a README, what code blocks look like, where commenting appears and what it says, etc.

I'm adding ad hoc comments since we have no such standards yet, but if we learn any lessons from this review it'd nice to keep in mind which lessons we'd like to generalize more broadly.

flags-parsing-tutorial/README.md
aranguyen: Added. Thanks for the link!

flags-parsing-tutorial/README.md
aranguyen: Done

flags-parsing-tutorial/README.md
aranguyen: The second one is a bit better. Thanks!

flags-parsing-tutorial/README.md
aranguyen: Linked

flags-parsing-tutorial/bazelrc
aranguyen: +1. I will do the same for other comment lines as well.

flags-parsing-tutorial/README.md
aranguyen: Thank you!

flags-parsing-tutorial/README.md
gregestren: Is that in your latest patch? I'm not sure why but I still see the old link.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,62 @@
+Bazel flags parsing examples
+========================
+
+This provides examples for the current behavior of flag/options parsing. This tutorial assumes that users have basic knowledge of creating a Bazel [WORKSPACE](https://docs.bazel.build/build-ref.html#workspace) and writing a [BUILD](https://docs.bazel.build/versions/main/build-ref.html#BUILD_files) file. Users should also know what [options/flags](https://docs.bazel.build/versions/main/command-line-reference.html#option-syntax) including built-in (non-Starlark) and [user-defined](https://docs.bazel.build/skylark/config.html#user-defined-build-settings) (Starlark) options are.
```
gregestren, line 4: You're not responsible for the overloading, but I wonder if we can figure out a better way to not have to mix "flags / options" all the time.

I agree that these are different concepts: an "option" implies anything in a BuildOptions while a flag is something set at the command line. https://docs.bazel.build/versions/main/skylark/config.html#the-build_setting-rule-parameter clarifies when it discusses the `flag` parameter.

But it's hard enough for us to keep the distinction straight and I can only imagine the conceptual overload casual users could face. 

These examples seem to be focused specifically on flags. Maybe we could commit to one term throughout the doc, then when the term is first mentioned have a quick clarifer that  the other term is a popular usage? Or even not have the clarifier at all?

flags-parsing-tutorial/README.md
aranguyen: I added a terminologies section before A and B instead  since we already mentioned --config in A

flags-parsing-tutorial/.ijwb/.bazelproject
```diff
@@ -0,0 +1,18 @@
+directories:
```
gregestren, line 1: Is this file intended for this PR?

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,101 @@
+Bazel flags parsing examples
+========================
+
+This provides examples for the current behavior of flag parsing. This tutorial assumes that users have basic knowledge of creating a Bazel [WORKSPACE](https://docs.bazel.build/build-ref.html#workspace) and writing a [BUILD](https://docs.bazel.build/versions/main/build-ref.html#BUILD_files) file. Users should also know what flags including [built-in](https://docs.bazel.build/versions/main/configurable-attributes.html#built-in-flags) (non-Starlark) and [user-defined](https://docs.bazel.build/versions/main/configurable-attributes.html#custom-flags) (Starlark) flags are.
```
gregestren, line 4: Optional wording nit: "Users should also be familiar with built-in (non-Starlark) and user-defined (Starlark) flags"? That would read a bit more concisely IMO.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,101 @@
+Bazel flags parsing examples
+========================
+
+This provides examples for the current behavior of flag parsing. This tutorial assumes that users have basic knowledge of creating a Bazel [WORKSPACE](https://docs.bazel.build/build-ref.html#workspace) and writing a [BUILD](https://docs.bazel.build/versions/main/build-ref.html#BUILD_files) file. Users should also know what flags including [built-in](https://docs.bazel.build/versions/main/configurable-attributes.html#built-in-flags) (non-Starlark) and [user-defined](https://docs.bazel.build/versions/main/configurable-attributes.html#custom-flags) (Starlark) flags are. 
+
+Note that a broader term, [options](https://docs.bazel.build/versions/main/command-line-reference.html#option-syntax) is often used interchangeably with flags. An important distinction is that only flags can be set on the command line. 
+
+Terminologies
+========================
+`--config`: Throughout this tutorial, users will see regular usage of `--config`. Although, it's already defined [here](https://docs.bazel.build/guide.html#bazelrc), important points will be repeated for first-time Bazel users.
```
gregestren, line 10: "we'll repeat important points for..."?

That's taking into account the guidance at https://bazel.build/basics/docs-style.html#tense.

flags-parsing-tutorial/README.md
```diff
@@ -0,0 +1,101 @@
...
+* `--config` can be used to represent a group of flags with a short name following the convention `<command>:<config_name>`. For example:
+```
+# bazelrc
+build:foo --//:wibble=wibble
+build:foo --//:wobble=wobble
+build:foo --//:wubble=wubble
+```
+Here, we have defined a `--config` named `foo` expanding flags `--//:wibble, --//:wobble, --//:wubble`.
+
+By default, flags defined in a `--config` will be ignored, unless `--config=<config_name>` is specified either on the command line or in the `bazelrc` file. Where `--config=<config_name>` is specified determines the priorities/precedence of the flag expanded by that `--config` (i.e RC priority vs command line priority).
```
gregestren, line 20: Same here.

gregestren: Also, I'm still having trouble parsing the last sentence. Is it saying that where you write `--config=foo` at the command line affects which priority it takes? Can we word in a way to avoid technical terms like "RC priority' and "command line priority"? I don't think casual users will understand them.

flags-parsing-tutorial/.ijwb/.bazelproject
aranguyen: no it shouldn't be here. I will remove

flags-parsing-tutorial/README.md
aranguyen: fixed

flags-parsing-tutorial/README.md
aranguyen: fixed. Thank you for the reference!

flags-parsing-tutorial/README.md
aranguyen: Yes, where `--config=foo` is written on the command line affects the priorities of the flags expanded by --config=foo. I agree with your thought, those details are probably not necessary here. Instead of that last sentence, how about _**"Flags grouped by --config are expanded in place having the same priority with the --config expanding them. Please see example B.4"**_?

flags-parsing-tutorial/README.md
aranguyen: We discussed this offline and the change for using flags throughout + a quick clarifier for options are reflected in the latest commit. Thank you for all the suggestions!

flags-parsing-tutorial/README.md
gregestren: Yes, that sounds great.

aranguyen: updated. Thanks!

flags-parsing-tutorial/README.md
aranguyen: change is reflected in latest commit.

gregestren (approved):

PR #231, opened 2021-12-13
adding example for --enable_platform_specific_config

aranguyen:


flags-parsing-tutorial/README.md
```diff
@@ -97,5 +97,20 @@ Commands such as test and release inherit flags from build. The inheriting comma
...
+build:macos --cpu=k8
+build:windows --cpu=x64_windows
+build:freebsd --cpu=ppc
+build:openbsd --cpu=haswell
+```
+If the host platform (where Bazel is running) is `macos` and the `build` command is run, Bazel picks up `build:macos` lines in the `bazelrc`. In this example, `build:macos --cpu=k8` will be enabled. Try the following command and observe the output, `k8` should appear in the output path.
+```
+bazel --bazelrc=./bazelrc build :wibble
+```
+Note that Bazel will only enable flags based on the host platform, instead of execution platform or target platform.
```
gregestren, line 115: Let's linkify probably to https://docs.bazel.build/platforms.html, since I'm sure causal users will get confused about what host / execution / target mean.

flags-parsing-tutorial/README.md
```diff
@@ -97,5 +97,20 @@ Commands such as test and release inherit flags from build. The inheriting comma
...
+```
+build --enable_platform_specific_config
 
+build:linux --cpu=arm
+build:macos --cpu=k8
+build:windows --cpu=x64_windows
+build:freebsd --cpu=ppc
+build:openbsd --cpu=haswell
+```
+If the host platform (where Bazel is running) is `macos` and the `build` command is run, Bazel picks up `build:macos` lines in the `bazelrc`. In this example, `build:macos --cpu=k8` will be enabled. Try the following command and observe the output, `k8` should appear in the output path.
```
gregestren, line 111: how about a parentheses after "output path" that explicitly says (`bazel-out/k8-fastbuild/...`), so beginner users know *exactly* what they should be looking for?

flags-parsing-tutorial/README.md
aranguyen: Done!

gregestren (approved):

PR #232, opened 2021-12-15
Creating example tests

KilimAnnejaro:
Per https://github.com/bazelbuild/bazel/pull/14432#issuecomment-994278938, adding example testing code to the sample repo.

@aiuto not sure if this is what you were looking for?

### DIFF
--- a/java-tutorial/BUILD
+++ b/java-tutorial/BUILD
@@ -15,7 +15,7 @@ java_library(
 java_test(
     name = "runner_tests",
     srcs = glob(["src/test/java/com/example/*.java"]),
-    test_class = "com.example.RunnerTest",
+    test_class = "com.example.CalculatorTest",
     deps = [
         ":java-maven-lib",
         "@maven//:junit_junit",
--- a/java-tutorial/BUILD
+++ b/java-tutorial/BUILD
@@ -1,6 +1,35 @@
-load("@rules_java//java:defs.bzl", "java_binary")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 java_binary(
     name = "ProjectRunner",
+    srcs = ["src/main/java/com/example/ProjectRunner.java"],
+    main_class = "com.example.ProjectRunner",
+    deps = [":greeter",":calculator"],
+)
+
+java_library(
+    name = "java-maven-lib",
     srcs = glob(["src/main/java/com/example/*.java"]),
+)
+
+java_test(
+    name = "runner_tests",
+    srcs = glob(["src/test/java/com/example/*.java"]),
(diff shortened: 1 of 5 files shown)


KilimAnnejaro:
@aiuto @brandjon @c-parsons bump?

java-tutorial/BUILD
```diff
@@ -1,6 +1,35 @@
...
 
 java_binary(
     name = "ProjectRunner",
+    srcs = ["src/main/java/com/example/ProjectRunner.java"],
+    main_class = "com.example.ProjectRunner",
+    deps = [":greeter",":calculator"],
+)
+
+java_library(
+    name = "java-maven-lib",
```
aiuto, line 11: Why java-*maven*-lib?  Isn't it just calculator_lib?

java-tutorial/BUILD
```diff
@@ -1,6 +1,35 @@
...
     srcs = glob(["src/main/java/com/example/*.java"]),
```
aiuto, line 12: Doesn't this have to exclude ProjectRunner.java?

aiuto (commented):
I'm going to pass this to someone who knows more about maven than I do. I never go near it if I can avoid it.

aiuto:
Do we have a standard way to include junit without having to bring in maven
as well?
That would make a better example because it does not conflate java_test
with maven.

java-tutorial/BUILD
```diff
@@ -1,6 +1,35 @@
...
+    test_class = "com.example.CalculatorTest",
+    deps = [
+        ":java-maven-lib",
+        "@maven//:junit_junit",
+    ],
+)
+
+java_library(
+    name = "greeter",
+    srcs = ["src/main/java/com/example/Greeting.java"],
```
jin, line 27: This doesn't exist..?

jin (changes requested):
Thanks for your contribution!

This repo is used with https://docs.bazel.build/versions/main/tutorial/java.html. The BUILD file doesn't have the greeter target for that reason, because the instructions wants the user to add it manually. So I don't think we should change the contents of this repo at all, unless there's an intention to update the java tutorial on docs.bazel.build in tandem.

Also, there are already Junit + rules_jvm_external test examples in https://github.com/bazelbuild/examples/tree/main/java-maven (see https://github.com/bazelbuild/examples/blob/d1a1f1fcf460a7f3aa0eb845b22dab061eff8ec9/java-maven/BUILD#L17), so I'm not sure what this PR is adding.

aiuto:
Going back to Jin's comment. This is part of the tutorial, so we don't want to change it without changing the tutorial in concert. I'm going to close this because it is so stale and someone can start again if they want to overhaul the whole tutorial.

PR #233, opened 2021-12-16
Set specific bazel version number for "Bazel Rules" job

meteorcloudy:
This will skip testing those jobs with Bazel@HEAD in downstream pipeline.
We can re-enable them in downstream when https://github.com/bazelbuild/bazel/issues/14361 is fixed.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -137,7 +137,7 @@ tasks:
   rules-linux:
     name: "Bazel Rules"
     platform: ubuntu1804
-    bazel: "4.2.2"
+    bazel: "latest"
     working_directory: rules
     build_targets:
     - "..."
@@ -155,7 +155,7 @@ tasks:
   rules-macos:
     name: "Bazel Rules"
     platform: macos
-    bazel: "4.2.2"    
+    bazel: "latest"    
     working_directory: rules
     build_targets:
     - "..."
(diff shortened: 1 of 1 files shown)


.bazelci/presubmit.yml
```diff
@@ -154,6 +155,7 @@ tasks:
   rules-macos:
     name: "Bazel Rules"
     platform: macos
+    bazel: "4.2.2"
```
aiuto, line 158: Do you want 4.2.2 or "latest" so it tracks 4.x releases?

aiuto (commented):
You might consider making it easy to define tests for both LTS and rolling. 
I do something like that here: https://github.com/bazelbuild/rules_pkg/blob/main/.bazelci/tests.yml

.bazelci/presubmit.yml
meteorcloudy: Oh, "latest" might be slightly better, since the breakage is not even in 5.0.

meteorcloudy: Done, please take another look.

meteorcloudy:
Ping, this is still blocking our downstream from turning green:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2290#d8ce930c-9daa-482a-a030-7ac9cdb19c39

meteorcloudy:
Can we get this merged so that I can reenable the examples project in Bazel's downstream

c-parsons (approved):

PR #237, opened 2022-01-19
Move in ios tutorial

aiuto:
Copy the iOS tutorial from the main bazel docs into this repository.

A follow up PR will remove https://docs.bazel.build/versions/main/tutorial/ios-app.html and point this new page instead.

Note that the tutorial is still a messy hodgepodge of stuff. The concept of the main vs source-only tree has not been maintained.  The tutorial looks like it is a pure iOS app, but the examples we point you to are mixed appengine, java, ios.

Someone should clean this up. It will be much easier when the example code and the tutorial instructions are in the same repository.

### DIFF
--- a/tutorial/ios-app/WORKSPACE.sample
+++ b/tutorial/ios-app/WORKSPACE.sample
@@ -1,0 +1,25 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "build_bazel_rules_apple",
+    remote = "https://github.com/bazelbuild/rules_apple.git",
+    tag = "0.32.0",
+)
+
+git_repository(
+    name = "build_bazel_rules_swift",
+    remote = "https://github.com/bazelbuild/rules_swift.git",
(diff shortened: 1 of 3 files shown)


aiuto:
@keith @skofgar
This incorporates the changes from https://github.com/bazelbuild/bazel/pull/14499.

keith (approved):

gregestren:
What's the state of this PR? I just had a ping on it but didn't realize it's been around for a long time now...

tutorial/ios-app/WORKSPACE.sample
```diff
@@ -0,0 +1,25 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
```
alexeagle, line 3: I don't think we should recommend git_repository since it has no entry in the Repository Cache.

Let's replace these with http_archive and also get the latest version?

alexeagle:
@cgrindel maybe this is a good place to start

PR #239, opened 2022-03-30
Fix markdown link in README.txt

swegner:
The line break between the `[configure attribute]` and the (link) itself broke markdown formatting.


gregestren (approved):
Nice catch!

aiuto:
@swegner Can you merge from upstream and push again.

PR #241, opened 2022-04-19
split yml file to distinct sections, stop building the broken examples

aiuto:
This just stops the pain. We can fix the examples later on.


jdob (approved):
LGTM

PR #242, opened 2022-04-19
Clean up the rules examples, part 1

aiuto:
Clean up the examples for the basics of rule writing.

1. A final goal would be to rename this to something more informative,
like "basic_rules", or "rules/basics". We are not doing that in this
PR because it will break all the pointers in docs.

Instead, this change starts to enable that by changing the absolute
references to targets within the examples to relative ones. That makes
it easier for people to grab the example and copy it into their own
code while having it still work.

2. Move starlark_configurations out of rules and into 'configurations'.
Someone should have learned the basics first before moving into
advanced topics.  I don't actually care about the final name for this,
because the short term goal is to get it out of the rules tree so
that I we can unblock CI for rules while trying to fix this. My only
concern for the name is that we don't call it 'starlark_configuration'
because that name only makes sense from the historicial point of view
of the bazel team, where we had native configuration and transitions.
From most users' point of view.  This is simply about defininging the
build configuration. It happens to be done on starlark because that
is the configuration language.

3. Rewrite the rules test configuration so it is a little easier to understand
and re-enable testing of the basic examples.


gregestren:
>  My only concern for the name is that we don't call it 'starlark_configuration'

Agreed. I use the term "user-defined" instead: "user-defined build flags", "user-defined transitions", etc.

gregestren (approved):

gregestren:
Conflicts for merging?

PR #264, opened 2022-11-11
Fix parsing book name from repo name

meteorcloudy:


Wyverald (approved):
this should ideally be fixed by passing in the book name explicitly as an attribute, instead of parsing it out of the canonical repo name; see https://github.com/bazelbuild/rules_java/commit/9b5ea1128461779f74ed3306e81d41872ff74bd4

PR #270, opened 2022-11-17
Remove @brandjon from CODEOWNERS

brandjon:
I get added automatically on reviews, but in truth I haven't actively looked at this repo in a long time. Any activity on my part in the near future is likely to be as a contributor rather than a maintainer.

This leaves Greg as the sole fallback though -- maybe we need to find a different owner for the root path?


gregestren:
I'll take the brunt for now. It's not high volume.

gregestren (approved):

PR #275, opened 2023-02-10
Update NDK example to use the pre-alpha rules_android.

katre:
Also use the rules_android_ndk repo.


754715:
> Also use the rules_android_ndk repo.

PR #282, opened 2023-03-25
Add an example of generating code and consuming it as a library.

aiuto:
This is a toy example that shows
- a rule that builds a binary that can generate some code from an input file
- using that rule to create python and C++ source files
- which are consumed by *_library rules
- and, finally, by *_binary rules.

### DIFF
--- a/rules/generating_code/BUILD
+++ b/rules/generating_code/BUILD
@@ -1,0 +1,32 @@
+
+load("//generating_code/gen:enum_maker.bzl", "enum_maker")
+
+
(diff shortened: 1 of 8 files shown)


tetromino (approved):
LGTM. I took the liberty of fixing a typo in the comment string and a pytype failure on macOS.

gregestren:
Always happy to see new examples!

Was this inspired by a particular use case? I'm asking because I'd like to have some rough taxonomy of the categories we want to document and why.

PR #283, opened 2023-04-03
Add a first-cut index to the README

aiuto:
- I tried to make the topic titles most likely to get search hits. So they might include extra words beyond the precise topic name.
- I excluded examples which were just confusing or wrong. TBH, some of the them are terrible. They are one line examples of a technique which should have been combined into a larger example that show two constrasting ways of doing something and their tradeoff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,5 +2,52 @@
 
 A collection of examples of how to use the Bazel build system.
 
+## Introductory tutorials
+
+* [C++ basics](/tree/main/cpp-tutorial)
+* [Java basics](/tree/main/java-tutorial)
+* [End to End](/tree/main/tutorial)
+  A full end-to-end system with a backend, an Android app, and an iOS app.
+* [iOS basics](/tree/main/tutorial/ios-app)
+* [Using query](/tree/main/query-quickstart)
+  Working examples for [The Query quickstart](https://bazel.build/query/quickstart)
+
+## Example cookbook
+
+### General
+
+* [Using bzlmod](/tree/main/bzlmod)
+* [How to generate source code from a rule and include it in a library](/tree/main/rules/generating_code)
+
+### [Android](/tree/main/android)
+
+* [Using Android Jetpack Compose](/tree/main/firebase-cloud-messaging)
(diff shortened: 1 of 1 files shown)


aiuto:
Notes:
- the links do not work until the PR is submitted because site relative links are borked in git review
- CI is broken for other reasons.

gregestren:
This is great - fantastically more useful landing page.

Any thoughts on keeping it in sync with the actual directories? In a more ambitious timeline it'd be cool if this could be auto-generated from annotations in the individual directories. 

Other thoughts:

- Which directories are excluded? Casual readers may assume examples don't exist if they're not explicitly part of the index. I see your point that you're intentionally excluding bad examples. But it'd be nice to have *some* pathway toward everything ultimately being in the index, or clearly marked otherwise.
- What's the story on the broken CI? Let's try to get CI generally green on this repo. I'm happy to help look into that further. We've already excluded some directories from the CI because they had hard-to-fix issues.

aiuto:
I have no good answer to the question of what is include because the state of the examples is in such bad shape.  I could not figure out what some of them even mean.
And some look like they are a multi-part example.  I just went with the ones that obviously made sense.    The path to improving the rest is for people to review things one by one and make sense of them.

The broken CI might have been mostly temporary.  I think xcode changed an all the machines had not updated yet.  Enough retries over the course of a day and then things worked.  OTOH.... CI is still flaky when things don't download. Many of the workspaces point to github rather than the mirror.

gregestren:
> I have no good answer to the question of what is include because the state of the examples is in such bad shape. I could not figure out what some of them even mean. And some look like they are a multi-part example. I just went with the ones that obviously made sense. The path to improving the rest is for people to review things one by one and make sense of them.
> 
> The broken CI might have been mostly temporary. I think xcode changed an all the machines had not updated yet. Enough retries over the course of a day and then things worked. OTOH.... CI is still flaky when things don't download. Many of the workspaces point to github rather than the mirror.

Got it - thanks for elaborating.

This is a great improvement over the status quo, so approving for that reason alone.

gregestren (approved):

PR #285, opened 2023-04-26
Update README.md

CanftIn:
fix relatively link of this pr: https://github.com/bazelbuild/examples/pull/283#issue-1652405365

maybe no neeed to use `/tree/main/`, github URL of completion is `https://github.com/bazelbuild/examples/blob/main/tree/main/cpp-tutorial`, this link will get 404:
image

should be `https://github.com/bazelbuild/examples/blob/main/cpp-tutorial`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,49 +4,49 @@ A collection of examples of how to use the Bazel build system.
 
 ## Introductory tutorials
 
-* [C++ basics](/tree/main/cpp-tutorial)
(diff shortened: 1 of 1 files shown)


gregestren:
@aiuto

gregestren:
Nice catch. @aiuto do you remember why the original links looked as they did?

I also wonder why the CI presubmit is failing?

aiuto (approved):

PR #307, opened 2023-09-06
Fix CI for C++ toolchain resolution

comius:
Apple rules are supported using `platform_mappings`. Unfortunately we need a copy, and cannot refer directly to the one provider by rules_apple. By default Bazel is looking for the file in root directory (so it's enough to just add the file).

The change passes CI with C++ toolchain resolution enabled on mac.


keith (approved):

comius:
cc @meteorcloudy

meteorcloudy (approved):

PR #308, opened 2023-09-22
chore: update bzlmod example

alexeagle:
- set the bazel version rather than recommend 'floating' last green
- recommend enabling bzlmod in .bazelrc
- simplify 'try it' commands
- MODULE.bazel.lock files now produced since it's bazel 6.3


gregestren:
Deferring to @Wyverald but let me know if you need permissions help with the `.bazelci/` diff.

bzlmod/05-integrate_third_party_package_manager/README.md
```diff
@@ -15,9 +15,8 @@ It covers the following topics:
 To test it out, `cd` into this directory and run the following:
 
 ```bash
-export USE_BAZEL_VERSION=last_green
-bazelisk build --enable_bzlmod //:check_books
-cat ./bazel-bin/books
+bazel build check_books
+cat $(bazel 2>/dev/null cquery --output=files check_books)
```
Wyverald, line 19: what's the reason for this change? it seems that using the bazel-bin path is much clearer?

Wyverald (approved):

bzlmod/05-integrate_third_party_package_manager/README.md
alexeagle: It's arbitrary, was just looking for another spot to illustrate that `cquery` allows you to write this kind of script without having to know the output path ahead of time (and accounts for `--symlink_prefix` being something else)

alexeagle:
I'll remove MODULE.bazel.lock files as well, since there are staleness bugs with bazel 6.3.0 per @fmeum

PR #314, opened 2023-09-27
chore: refactor third-party-dependencies for bzlmod

alexeagle:
Use a simpler pattern which I've seen used at every one of Aspect's clients. I've never seen the more complex pattern that was described here.


meteorcloudy (approved):
Thanks!

PR #350, opened 2023-10-30
chore(bzlmod): enable bzlmod in android/ndk

alexeagle:


alexeagle:
@Bencodes @ahumesky I've got an assignment from the Bazel team to upgrade the workspaces in this repo to use bzlmod, but I'm not sure if Android is ready for this yet or not?

PR #355, opened 2023-10-31
feat: add gcc toolchain example

f0rmiga:


alexeagle:
@f0rmiga do you have a suggestion where I should start to green this up by getting mac and windows to fall through to the system toolchain?

f0rmiga:
> @f0rmiga do you have a suggestion where I should start to green this up by getting mac and windows to fall through to the system toolchain?

I can take that up later tonight and make it green!

alexeagle (approved):
Nice, thank you!

Wyverald:
Bit late to the party, but I'm trying to migrate the C++ tutorial to use Bzlmod. Are there plans to make `aspect_gcc_toolchain` a module?

In any case, I feel like the `stage0` here isn't really what a first-timer wants to see. Sure it's useful to set up a hermetic toolchain, or to know how to cross-compile, but that's really advanced stuff considering `stage1` is literally a "hello world".

I'd rather we revert this change and put the gcc toolchain in a separate tutorial instead (or maybe as a `stage4`). WDYT? @alexeagle @f0rmiga

alexeagle:
Agree it doesn't have to be hermetic from the start, and that's not the typical developer path today. I can add stage4 with the llvm toolchain.

PR #367, opened 2023-11-23
Add `cfg = _test_arg_transition` in configurations/cc_test/defs.bzl

rwgk:
Question for the bazel experts: Does configurations/cc_test work as intended without the added `cfg = _test_arg_transition` in defs.bzl?

This PR changes mytest.cc to send `argv` to stdout.

Without the added `cfg = _test_arg_transition` in defs.bzl, `new arg` does not appear in the output. See below.

The other changes are optional, they just make this example more complete. Not being very familiar with the bazel mechanisms, it took me a while to figure out the two tricks to 1. ensure that `args` are handled correctly, and 2. the non-transitioned test is not run.

________

**WITHOUT** the added `cfg = _test_arg_transition` in defs.bzl

```
bazel test :all --test_output=all
```
```
MYTEST ARGV[0]: /usr/local/google/home/rwgk/.cache/bazel/_bazel_rwgk/bfd421ff56c7e52ce2de56579817e14d/sandbox/linux-sandbox/9/execroot/__main__/bazel-out/k8-fastbuild/bin/my-test.runfiles/__main__/my-test
MYTEST ARGV[1]: x
MYTEST ARGV[2]: y
MYTEST ARGV[3]: z
```
________

**WITH** the added `cfg = _test_arg_transition` in defs.bzl (i.e. this PR as is)

```
MYTEST ARGV[0]: /usr/local/google/home/rwgk/.cache/bazel/_bazel_rwgk/bfd421ff56c7e52ce2de56579817e14d/sandbox/linux-sandbox/4/execroot/__main__/bazel-out/k8-fastbuild-ST-54535d7cadf4/bin/my-test.runfiles/__main__/my-test
MYTEST ARGV[1]: x
MYTEST ARGV[2]: y
MYTEST ARGV[3]: z
MYTEST ARGV[4]: new arg
```


gregestren:
Hi @rwgk .

Thanks for your contribution.

I agree the existing test is sparse. And adding more examples of using the idea makes the point more clear. So I like what you're doing.

The main impact of  transitioning directly on the `transition_rule_test` as you're doing is the final output may no longer appear under `bazel-bin`. That's arguably not important if we're just running a test. See for example https://github.com/bazelbuild/bazel/pull/18854. 

Who produces `new arg` in your example?

configurations/cc_test/defs.bzl
```diff
@@ -26,6 +26,7 @@ def _test_transition_rule_impl(ctx):
     return [DefaultInfo(runfiles = runfiles, executable = executable_dst)]
 
 transition_rule_test = rule(
+    cfg = _test_arg_transition,
     implementation = _test_transition_rule_impl,
     attrs = {
         "actual_test": attr.label(cfg = _test_arg_transition, executable = True),
```
gregestren, line 32: Whatever we do we don't want to apply the transition twice. In this case it's just redundant but it's a pattern I'd discourage.

gregestren (commented):
I think overall this s a nice improvement, and makes the example more complete as you describe.

There are all kinds of subtle implications in transitions that users should be aware of. But this example doesn't have to explore the full space, and a more complete example is better than a less complete one. So I'm strongly inclined to support your improvements without worrying too much.

I'd also update the `README.md` to walk through what's going on more cleanly. As written I think it's overly vague.

configurations/cc_test/defs.bzl
rwgk: Thanks Greg! I need to find a moment (later) to explain more. For now just quickly:

I tried removing `cfg = _test_arg_transition` in the `attr.label()` call but then got an error reporting that the `cfg` argument is mandatory with `executable = True`. So I put it back.

What I was wondering about (with a huge amount of uncertainty): would it be better to use `cfg = "exec"`?

gregestren: Oh. If it's mandatory I'd set `cfg = "target"`, which is a no-op. If you search for 'cfg' on https://bazel.build/rules/lib/toplevel/attr it talks about this.

I think `cfg = "exec"` would mess with test targets too much. Tests are weird in that they involve both target and exec platform properties.

rwgk: The change to `cfg = "target"` is under commit 7a31decce5dbf901997fe33ffdef81aab9f636da.

rwgk:
> I'd also update the `README.md` to walk through what's going on more cleanly. As written I think it's overly vague.

Done (28024ad). I'm simply showing the bazel command and expected output. Please let me know if you had something else in mind.

gregestren (approved):

PR #369, opened 2023-12-03
Add react webpack example

Mivr:
Related to https://github.com/aspect-build/bazel-examples/pull/276 and https://github.com/aspect-build/bazel-examples/issues/83

This PR adds a react JS Webpack example using SWC for transpilation and Webpack for bundling.


alexeagle:
Looks great, thank you!

FYI I've been working the last couple days to change the existing react example to stop using react-scripts, but I'll switch it to vite/vitest so it will still be nicely orthogonal to this one.

PR #376, opened 2023-12-08
feat(react): use vite and vitest rather than react-scripts

alexeagle:
react-scripts (the build system stamped out by create-react-app aka CRA) is no longer recommended by https://react.dev/learn/start-a-new-react-project and instead developers are encouraged to use 'a framework'.
In the "Can I use React without a framework?" section, they suggest rolling up your sleeves and just use a bundler like Vite - that's what we do here.

This PR mostly prepared just by following public guides, such as 
https://dev.to/henriquejensen/migrating-from-create-react-app-to-vite-a-quick-and-easy-guide-5e72

Note that vitest just hit 1.0 three days ago (as of this writing) so it's good timing for us to demonstrate adoption. We got some confidence in this recommendation since Sourcegraph recently did it: https://github.com/sourcegraph/sourcegraph/pull/57886

### DIFF
--- a/frontend/react/vite.config.js
+++ b/frontend/react/vite.config.js
@@ -5,10 +5,7 @@ import svgr from 'vite-plugin-svgr'
 // https://vitejs.dev/config/
 export default defineConfig({
   base: '/',
-  plugins: [svgr({
-    // A minimatch pattern, or array of patterns, which specifies the files in the build the plugin should include.
-    include: "**/*.svg?react"
-  }), react()],
+  plugins: [svgr(), react()],
   test: {
     globals: true,
     css: true,


frontend/react/src/BUILD.bazel
```diff
@@ -1,92 +1,62 @@
...
 )
 
 ts_project(
-    name = "src_ts",
+    name = "src",
     srcs = glob(
         include = SRC_PATTERNS,
         exclude = TEST_PATTERNS,
     ),
+    data = [":assets"],
```
jbedard, line 19: Should we use `assets` instead?

alexeagle: It's just a js_library so the copy_to_bin happens, with a `filegroup` you get

```
Expected to find file App.css in //react, but instead it is in //react/src.

To use copy_to_bin, either move App.css to //react, or move the copy_to_bin
target to //react/src using:

    buildozer 'new copy_to_bin App' //react/src:__pkg__
    buildozer 'add srcs App.css' //react/src:App
    buildozer 'new_load @aspect_bazel_lib//lib:copy_to_bin.bzl copy_to_bin' //react/src:__pkg__
    buildozer 'add visibility //react:__subpackages__' //react/src:App
```

alexeagle: Note: I think the desirable answer is that this assets label wouldn't be used outside this package, they are already a data dep of the ts_library rule. But, the `vite` bundle action in the parent folder currently relies on a separate dep on the `assets` in order to resolve the css file.

jbedard: With `assets` you should get the copy_to_bin as well, but I guess you also need the copy for `vite`?

jbedard (approved):

Issue #381, opened 2023-12-11
fatal error: jni.h: No such file or directory with Bazel@HEAD


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3519#018c569e-3795-4226-ab3e-197736d4484f

Platform : Windows, Ubuntu, MacOS

Logs:
```
app/src/main/cpp/native-lib.cpp:1:10: fatal error: jni.h: No such file or directory
 #include <jni.h>
          ^~~~~~~
compilation terminated.
(02:09:02) ERROR: /var/lib/buildkite-agent/builds/bk-docker-l9gw/bazel-org-repo-root/examples/android/ndk/app/src/main/BUILD.bazel:17:11: Compiling app/src/main/cpp/native-lib.cpp failed: (Exit 1): gcc failed: error executing CppCompile command (from target //app/src/main:jni_lib)
```
Steps:
```
git clone -v https://github.com/bazelbuild/examples.git
git reset e2746bf0e3f987cc556e5188c18fec33edc1b61a --hard
export USE_BAZEL_VERSION=591d1257c51c2f997e067964878e3a9c7ee1f89c
cd /android/ndk
bazel build --noincompatible_enable_android_toolchain_resolution --fat_apk_cpu=arm64-v8a,x86 --android_crosstool_top=@androidndk//:toolchain --enable_bzlmod  //app/src/main:app
```
CC Greenteam @comius

meteorcloudy:
Looks like the failure is related to the changes in https://github.com/bazelbuild/examples/commit/5e37d3c0041dc8e4750f6dfbd7508782cdc3c610

/cc @katre Can you please take a look?

katre:
As indicated [in `android.yml`](https://github.com/bazelbuild/examples/blob/36408ec43bb419d483c30cd73e08a32dd7dfefe5/.bazelci/android.yml#L17), the problem is that this example does not use the new Android platforms flags, it's using the legacy flags, which aren't supported in Bazel at HEAD.

Using Bazel 6.4.0, the build works:
```
$ USE_BAZEL_VERSION=6.4.0 bazel build --noincompatible_enable_android_toolchain_resolution --fat_apk_cpu=arm64-v8a,x86 --android_crosstool_top=@androidndk//:toolchain --enable_bzlmod  //app/src/main:app
INFO: Analyzed target //app/src/main:app (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
[...]

PR #397, opened 2024-01-16
Disable some android example testing in Bazel downstream

meteorcloudy:
Due to https://github.com/bazelbuild/examples/issues/381


gregestren (commented):
Can we avoid this now thanks to https://github.com/bazelbuild/examples/pull/398 ?

meteorcloudy:
Yes, thanks for the fix!

PR #419, opened 2024-03-08
feat: add astro example

pedrobarco:
I've originally created [this ](https://github.com/pedrobarco/astro-bazel) repo to test it  and shared it in slack but @alexeagle made me do this 😎

Let me know if there's anything you'd like to see in this particular example.

I still need to understand how to fix the following warning:
```sh
WARNING: <repo>/examples/frontend/BUILD.bazel:10:22: input 'package' to //:.aspect_rules_js/node_modules/@swc+core@1.3.100/lc is a directory; dependency checking of directories is unsound
```


frontend/astro/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+
+SRCS = [
+    "package.json",
+    "tsconfig.json",
+    "astro.config.mjs",
+    "//astro/src",
+    "//astro/public",
+]
+
+BUILD_DEPS = [":node_modules/" + d for d in [
```
alexeagle, line 15: nit: do you expect to add more here? seems like this expression should be simplified

alexeagle (approved):

alexeagle:
From the CI failure
```
EPERM: operation not permitted, mkdir '/Users/buildkite/Library/Preferences/astro'
  Stack trace:
    at Object.mkdirSync (node:fs:1391:3)
    at get store [as store] (file:///private/var/tmp/_bazel_buildkite/a7160a15a89154e2ab8ab503e69ab355/sandbox/darwin-sandbox/2490/execroot/_main/bazel-out/darwin_x86_64-fastbuild/bin/node_modules/.aspect_rules_js/@astrojs+telemetry@3.0.4/node_modules/@astrojs/telemetry/dist/config.js:39:10)
```
it looks like the tool expects to modify the host machine filesystem, can it be configured not to do that?

frontend/astro/BUILD.bazel
pedrobarco: In this particular example I won't but astro provides a bunch of [integrations](https://docs.astro.build/en/guides/integrations-guide/) with UI frameworks, SSR adapters, mdx, tailwind, etc. So was expecting users to have a dedicated block for `BUILD_DEPS`

pedrobarco:
> From the CI failure
> 
> ```
> EPERM: operation not permitted, mkdir '/Users/buildkite/Library/Preferences/astro'
>   Stack trace:
>     at Object.mkdirSync (node:fs:1391:3)
>     at get store [as store] (file:///private/var/tmp/_bazel_buildkite/a7160a15a89154e2ab8ab503e69ab355/sandbox/darwin-sandbox/2490/execroot/_main/bazel-out/darwin_x86_64-fastbuild/bin/node_modules/.aspect_rules_js/@astrojs+telemetry@3.0.4/node_modules/@astrojs/telemetry/dist/config.js:39:10)
> ```
> 
> it looks like the tool expects to modify the host machine filesystem, can it be configured not to do that?

I've track it down to the way [telemetry is stored in astro's global config](https://github.com/withastro/astro/blob/dfdf6b3a0b50e14754abdbcc62fd714f80bf0666/packages/telemetry/src/config.ts#L44) and it's now working as expected. I've added the  `ASTRO_TELEMETRY_DISABLED=1` env var to disable collecting [telemetry](https://astro.build/telemetry/) data so that there's no need to create a global astro preferences dir.

alexeagle:
Looks like the node version you chose doesn't work on Ubuntu 18.04 - I think you need an older node to keep the glibc compat
'rules_nodejs~6.0.5~node~nodejs_linux_amd64/bin/nodejs/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found'

pedrobarco:
> Looks like the node version you chose doesn't work on Ubuntu 18.04 - I think you need an older node to keep the glibc compat
> 
> ```
> rules_nodejs~6.0.5~node~nodejs_linux_amd64/bin/nodejs/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found
> ```

I've tested a couple of versions and it looks like the last node version that uses the glibc version that comes with Ubuntu 18.04 is node 17, which unfortunately isn't supported by [astro](https://docs.astro.build/en/install/auto/#prerequisites) (v18.14.1 or higher).

```sh
root@c05651222d38:/# ldd --version
ldd (Ubuntu GLIBC 2.27-3ubuntu1.6) 2.27
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

```sh
root@c05651222d38:/# nvm use 17
Now using node v17.9.1 (npm v8.11.0)
root@c05651222d38:/# node --version
v17.9.1
```

```sh
root@c05651222d38:/# nvm use 18
Now using node v18.19.1
root@c05651222d38:/# node --version
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
```
```sh
root@c05651222d38:/# nvm use 20
Now using node v20.11.1
root@c05651222d38:/# node --version
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
```

alexeagle:
I don't control the machines that run the presubmits here, that's someone at Google. The config just has https://github.com/bazelbuild/examples/blob/main/.bazelci/frontend.yml which doesn't let us pick a newer ubuntu.

It's really lame that node doesn't ship a statically-compiled binary (link against musl, for example)...

alexeagle:
Rebased now that CI is functional again (thanks @meteorcloudy )

alexeagle (approved):
thank you!

PR #422, opened 2024-03-12
feat: add sveltekit example

yesudeep:
Adding a sveltekit example based on aspect's rules_ts and rules_js.


alexeagle:
It's actually our validator that's failing, you could set `validate = False` to bypass and see what error you get next.

If sveltekit really does need to generate a folder, you could either generate ahead of time like with `bazel run` and check the result into the project, or make an action that generates (js_run_binary with out_dirs) and then feed the result to this compilation step.

Sorry this will probably take someone's actual debug time to get deeper.

PR #430, opened 2024-03-20
refactor: rename frontend to javascript and upgrade to rules_js 1.39.1

gregmagolan:
Should fix the flake seen in the frontend example CI caused by https://github.com/aspect-build/rules_js/issues/1412.

`javascript` is a better name for the folder IMO but the rename is also related to fixing the flake since it results in all repository rules in that directory being invalidated and re-run. The above flake is tied to when the repository rule executes and on persistent runners on Bazel CI just bumping the rules_js version is not sufficient to make that happen.


alexeagle:
I think renaming this is too disruptive, the examples repo is VERY commonly used as a reference which means this rename breaks a lot of folks permalinks, in their "getting started" docs.

PR #449, opened 2024-05-01
Update CODEOWNERS

alexeagle:
It's not wise to have a single person as the root reviewer, in case they are unavailable for some reason.

Also replace Sven with Yun, and add Florian for CI changes

FYI @fweikert if you could nudge Greg to get this in :)


gregestren (approved):

gregestren:
Welcome @Wyverald !

PR #518, opened 2024-10-10
Create example for usage of Symbolic Macros

brandjon:
Greg and Xudong are optional reviewers, you were added automatically as code owners. Sasha and Susan are familiar with the content and it's their review (or rubber stamp) I'm seeking.

This adds a new directory containing a set of interrelated symbolic macros that builds a narrative in inline comments. It's intended as reference material to be linked to from the official bazel docs. It explains core symbolic macro features, including key aspects of the visibility system, and has a brief example of a finalizer.

Work toward bazelbuild/bazel#23857.


macros/README.md
```diff
@@ -0,0 +1,9 @@
+# Macros
```
brandjon, line 1: Note to self: Add mention that these examples require a non-hermetic `python3` be available in the genrule environment.

tetromino:
(I took the liberty of fixing buildifier formatting issues)

macros/fancy_word_counter/defs.bzl
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+    count_words(
+        name = name,
+        visibility = visibility,
+        # If we're forwarding our internal <name>_shouted target, count_words
+        # is able to instantiate dependencies on it because we passed the label
+        # in here as an attribute, even though package //word_counter is not in
+        # <name>_shouted's visibility.
+        srcs = [name + "_shouted"] if make_it_loud else srcs,
```
tetromino, line 32: Add a comment that `if make_it_loud` is valid only because make_it_loud is nonconfigurable?

tetromino (approved):

tetromino:
I might suggest is adding a diff test to verify the macros' output, but can postpone to after the 8.0 release, since we can't run CI on macro examples until 8.0

macros/README.md
brandjon: Done.

macros/fancy_word_counter/defs.bzl
brandjon: I think it's enough to mention it at the first use of `make_it_loud` above. Though instead of saying it's not valid, since that behavior's not implemented yet, I just described the existing silent-pass behavior and linked to the tracking bug.

brandjon:
> (I took the liberty of fixing buildifier formatting issues)

Thanks!

brandjon:
I added a comment about select() promotion because I thought it was useful context on why people should want this feature. It might be a little bit out of place in a reference example as opposed to a presentation or email advocating symbolic macros. But I thought it might be handy when users are wondering why they have to deal with select() so much when using symbolic macros.

macros/word_counter/defs.bzl
```diff
@@ -0,0 +1,66 @@
...
+        # The value of our visibility parameter is whatever was passed in by
+        # our caller, plus the location of the call site (so that our exported
+        # targets are guaranteed to always be visible to the caller without
+        # them having to explicitly say so).
+        #
+        # We do *not* write visibility = ["//visibility:public"], which would
+        # make this target exposed to the entire build, ignoring the visibility
+        # passed in from the caller.
+        visibility = visibility,
+        outs = [out],
```
susinmotion, line 20: It's a little confusing that we enforce the naming schema, but we make it easy to violate it. Do you think it would make sense not to take the "out" parameter?

susinmotion (approved):

macros/word_counter/defs.bzl
brandjon: I guess I went with an explicit output to more closely match `genrule`. I'd rather not rework the example at this point, but I'll elaborate the comment on the `out` param of target `//main:hamlet`.

PR #529, opened 2024-11-27
New frontend example for react-router

jimmyt857:
With the recent merge between React Router and Remix, React Router v7 is a compelling option as a frontend framework and I wanted to try it out. Sending a PR because I thought others might benefit from my attempt at getting this working within Bazel!

Although the dev and production versions of the server are both working, this example is 95% of the way there. Unfortunately I'm not yet deeply familiar with `rules_js`, `rules_ts`, or even `react-router`. I need a bit of help fixing the `//react-router/app:app_typecheck` and `//react-router/app:test` tests, they are each failing for a different reason:

- `:app_typecheck` fails because it is missing some special typing files generated by `react-router`. See their [documentation here](https://reactrouter.com/explanation/type-safety). I already wrapped the `react-router typegen` command into a `js_run_binary` but I couldn't figure out how to make the `ts_library` target accept the generated files as input.
- `:test` fails because the `vite` library somehow doesn't contain `expect`. Honestly not sure why this is happening, the test code is taken almost exactly from https://github.com/bazelbuild/examples/blob/main/frontend/react/src/App.test.tsx.


frontend/react-router/app/BUILD.bazel
```diff
@@ -0,0 +1,71 @@
...
+        allow_empty = True,
+    ),
+    declaration = True,
+    resolve_json_module = True,
+    transpiler = TRANSPILER,
+    tsconfig = "//react:tsconfig",
+    deps = [
+        ":app",
+        "//:node_modules/vitest",
+        "//react-router:node_modules/@testing-library/jest-dom",
```
thesayyn, line 61: These don't have to be absolute, `:node_modules/@testing-library/jest-dom` would work as well, and actually prefer relative labels as it saves me the brain cycles  "where this label refer to, and what folder this BUILD file is in" vs "oh it's relative, somewhere within my turf". 

That however is your choice, perhaps you already adopted some rules for absolute labels.

frontend/react-router/app/entry.server.tsx
```diff
@@ -0,0 +1,75 @@
+// This file is copied from https://github.com/remix-run/react-router/tree/main/packages/react-router-dev/config/defaults.
+//
+// Normally, this file is optional (https://reactrouter.com/explanation/special-files#entryservertsx) and react-router would use the copy in its own package as the default.
+// However, when run in Bazel it is not able to find the default copy at runtime.
+// TODO: Figure out how to make this work in Bazel.
```
thesayyn, line 5: what's broken here?

thesayyn (approved):
It looks good to me.

alexeagle (commented):
I'll pull this down and try it out, maybe I can see what needs fixing and/or add some automated testing so the PR would be red

frontend/react-router/app/entry.server.tsx
jimmyt857: The "broken" behavior is the fact that this file must be vendored into our repo.

The default behavior of react router is to fallback to its own internal copy of this file, when this file is omitted. Somehow though, when running in bazel, the fallback fails. I'm copying the fallback file into my own repo so that react router can find it.

alexeagle:
I checked in rules_ts and actually we do have a very example for combining folders with rootDirs.
https://github.com/aspect-build/rules_ts/tree/main/examples/root_dirs
This can work here as well. The key is that the `ts_project` has to be at the common ancestor folder of the rootDirs, since we set `rootDir` to control output locations. For some reason this morning I'm getting a permission denied pushing to your branch so I've put a commit to illustrate over here: https://github.com/bazelbuild/examples/compare/main...alexeagle:examples:11-26-new_frontend_example_for_react-router

jimmyt857:
Thanks Alex! I've pulled in your code and also rebased this branch. I think the only remaining problem is that the unit test still fails, with an error saying that `expect` somehow cannot be imported from `vite`.

alexeagle (approved):

PR #536, opened 2025-01-02
Update NDK example to newest versions of rules

ted-xie:
Fixes #522


ahumesky (approved):

ahumesky:
looks like this may have rotted a bit, getting an error in CI for absl?

Issue #540, opened 2025-01-07
[Bazel CI] Error in glob: glob pattern 'lib/clang/*/lib/**/*.dylib' didn't match anything



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4381#01943ef5-b088-4c02-a5ba-2aa30cdd04f3

Platform: Ubuntu

Logs:

```
 ERROR: Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e5b85cc96123ab88632bda12b9f1b123/external/toolchains_llvm++llvm+llvm_toolchain_llvm/BUILD.bazel", line 60, column 16, in <toplevel>
		srcs = glob(
Error in glob: glob pattern 'lib/clang/*/lib/**/*.dylib' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
(04:14:01) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e5b85cc96123ab88632bda12b9f1b123/external/toolchains_llvm++llvm+llvm_toolchain/BUILD.bazel:114:10: Target '@@toolchains_llvm++llvm+llvm_toolchain_llvm//:clang' contains an error and its package is in error and referenced by '@@toolchains_llvm++llvm+llvm_toolchain//:compiler-components-x86_64-linux'
(04:14:01) ERROR: Analysis of target '//:all' failed; build aborted: Analysis failed
```

Steps:

```
git clone https://github.com/bazelbuild/examples && cd ../rust-examples/02-hello-cross
git reset b51e3bdd468ce8c4a516d7dca993909dcc84af32  --hard
export USE_BAZEL_VERSION=de4ba2f81c87eeb9a25b6a9777889785f6e1ed7a
bazel build //:all 
```

CC Greenteam @Wyverald

meteorcloudy:
@marvin-hansen Are you able to take a look at the test failures of the rust examples with Bazel@HEAD? See the latest build result from https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4431#_

If not, we'll have to skip those tests in bazel downstream pipeline.

marvin-hansen:
Okay, I take a look over the upcoming weekend.

marvin-hansen:
@meteorcloudy @sgowroji 

[...]

PR #546, opened 2025-02-02
Fix for Rust examples 1 - 3

marvin-hansen:
In response to issue https://github.com/bazelbuild/examples/issues/540#issuecomment-2621732201

I am working through all the Rust examples in smaller batches and filling PR's with at most 3 renovated or fixed exmples to keep the review overhead as little as possible. 

This one starts with example 1 - 3. 
Besides a number of small updates an fixes, I took the freedom to replace the normal LLVM clang compiler
from the cross example (02) with a much smaller mini version that is about 95% smaller in download size 
which hopefully cuts CI times to a minimum.


marvin-hansen:
Closed with https://github.com/bazelbuild/examples/pull/548

PR #548, opened 2025-02-02
Fix for Rust examples 8 - 9

marvin-hansen:
Last batch of fixed up Rust examples.
Example 8 got a major makeover b/c the latest rules_rust separated into multiple sub-modules,
which required quite some refactoring. 
Resolves https://github.com/bazelbuild/examples/issues/540


meteorcloudy:
The number of files for "07-deps-vendor" went up by a lot, is this absolutely necessary in the example?

marvin-hansen:
Well,

I can restructure the example to exclude vendoring, which cuts out all
those files.  The example still does the same just without the third
partyfolder.

If that's prefered, i push out a patch probably tomorrow.

marvin-hansen:
I just looked at the sample code and it's actually the vendoring example.


This is an ugly topic because remote vendoring has changed a lot. Let me
ask on the rules_rust slack to figure that one out. I'm only using local
vendoring for a number of reasons but it's time to figure local vendoring
out again.

marvin-hansen:
I actually tried native bzlmod vendoring once and after having downloaded
about 25GB of every Bazel rule for every platform under the sun, I decided
it's really only for total corner cases i.e. air gapped builds. In all
fairness, the official documentation says basically the same so it's really
only for those who need a total level of hermitic isolation. I would skip
this for the rust examples.

The tricky thing with rules_rust is, there are actually 4 different vendor
modes and I sense it's about high time to set the record straight and
document that correctly in the examples.

Let me think this through and update the example in a sensible way.

meteorcloudy:
> I actually tried native bzlmod vendoring once and after having downloaded about 25GB of every Bazel rule for every platform under the sun

😅, I guess that's before we implemented vendoring targets? With https://bazel.build/external/vendor#vendor-target-dependencies, the Bzlmod vendor mode will only vendor dependencies required to build your targets for the current configuration, which should be much less dependencies than vendoring the whole world.

marvin-hansen:
That's exactly right, I tried out the very first version that came out...

I give it another try when I find a bit of time because vendoring target
sounds about right to me...👍👍👍


Thanks for the info.

marvin-hansen:
@meteorcloudy 

Can I pick up your brain?

I had the idea to just remove the vendored folder from example 7, add the vendor target to CI, and update the Readme
with instructions how to vendor the depds before running the build. However, Bazel CI somehow doesn't support setting a run target in the config, but without it the CI task fails b/c just run bazel build //vendor:target doesn't create the folder with the vendored deps. For now, I have rolled things back so that CI is green again.

There are a few options:
1) One might get around this with some shell and gen rule hacking. Don't know, would have to investigate.
2) One can also just leave it as it and having ~1500 files in the example folder. 

The actual crux is, when I looked into Rust vendoring again, I realized I may have to split the example into 
4 different ones so I prefer to get this PR resolved, merged, and push the update of the vendor example into a seperate
PR b/c it's quite a bit of work on my end.

That means, whatever resolution for example 7 will be reached, it only lasts as long until I find the time to 
split the vendoring example so that it showcases all four vendor modes. Why the rules_rust team came up with the idea that anyone needs/wants 4 different ways to vendor on top of the existing Bazel vendor mechansim is a totally different discussion, but for now I advocate to get this PR over the finishing line and make the renovation of Rust vendoring a separate PR.

meteorcloudy:
Have you tried https://github.com/bazelbuild/continuous-integration?tab=readme-ov-file#running-commands-shell-scripts-or-binary-targets? Using either `shell_commands` or `run_targets`?

I'm really not a big fan of checking in 5k+ generated files if we don't have to.

marvin-hansen:
Perfect.

Let me try the run target and if that works, i just delete the vendored
folder. Didn't know the option exists so thanks for the docs.

marvin-hansen:
And that did the trick; 
just deleted all the generated files, configured CI to do the vendoring, and build is green. 
Thanks @meteorcloudy

meteorcloudy:
Nice! But I still see 5k+ changed? Looks like there are more files to delete?

marvin-hansen:
Yeap, fixed up Rust example 8 and 9.
Changed files down to ~200,  which is more reasonable.

meteorcloudy (approved):

PR #554, opened 2025-02-16
Fixes Rust example 8 failing on downstream Bazel Head CI

marvin-hansen:
This PR resolves half of the failing Rust examples on downstream Bazel built from HEAD.

The two examples in questions are 7 and 8.  
Example 7 is tricky and gets its own PR when I found a workaround. 

This PR resolves example 8 by removing the vendoring that causes the build failure on downstream.


marvin-hansen:
Resolves partially https://github.com/bazelbuild/examples/issues/540

meteorcloudy:
Why do we have so many file changes? Some vendored file not removed?

marvin-hansen:
Because in order to do the crate annotation
necessary for the proto toolchain without vendoring,
i had to rewrite the entire example. It's one of the weirder things in
rules rust_that you just can flip the vendor mode on and off when you work
with custom or proto toolchains because the dependency path in the
annotation changes.

Looking back, using vendoring in that specific example to simplify working
with proto was a clear mistake I made back when drafting the example.

meteorcloudy:
I still don't understand why there are so many new files, I'm not fine with checking in a large number of files in this repo. I won't be able to review them.

marvin-hansen:
Actually, you are right. There is too much unrelated stuf. Sorry, I'm on my mobile and couldn't really see it.

Let me move the changeset into a clean branch tomorrow and try again.

marvin-hansen:
Okay, @meteorcloudy I've filled a new PR with just the 7 affected files.
CI is already green. 

Please review: https://github.com/bazelbuild/examples/pull/555

Issue #562, opened 2025-03-10
[Bazel Examples CI] Error : No such target '//thirdparty/crates:tokio-1.43.0': target 'tokio-1.43.0' not declared in package 



sgowroji:
CI: https://buildkite.com/bazel/bazel-bazel-examples/builds/4189#01957e36-21cf-4b29-a044-72b76005dcd6

 Platform: Ubuntu

 Logs: 
```
ERROR: /workdir/rust-examples/07-deps-vendor/thirdparty/crates/BUILD.bazel: no such target '//thirdparty/crates:tokio-1.43.0': target 'tokio-1.43.0' not declared in package 'thirdparty/crates' defined by /workdir/rust-examples/07-deps-vendor/thirdparty/crates/BUILD.bazel (did you mean tokio-1.44.0?)
(04:02:16) ERROR: /workdir/rust-examples/07-deps-vendor/basic/BUILD.bazel:3:12: no such target '//thirdparty/crates:tokio-1.43.0': target 'tokio-1.43.0' not declared in package 'thirdparty/crates' defined by /workdir/rust-examples/07-deps-vendor/thirdparty/crates/BUILD.bazel (did you mean tokio-1.44.0?) and referenced by '//basic:hello_vendored'
(04:02:16) ERROR: Analysis of target '//basic:hello_vendored' failed; build aborted: Analysis failed
``` 

 CC Greenteam @meteorcloudy

meteorcloudy:
@marvin-hansen Can you take a look? Looks like the dependency resolution wasn't reproducible and now resolved to a newer version?

meteorcloudy:
The build was green at the same commit: https://buildkite.com/bazel/bazel-bazel-examples/builds/4174

marvin-hansen:
Yea,

I know this issue. Let me think through it because the root cause is
actually leaky vendoring in rules_rust i.e. version mismatch when the rules
download a higher version than the one that is used. I may have to add some
aliasing since rust crates usually don't break at minor version changes.

Takes probably a day or two.

marvin-hansen:
@meteorcloudy PR https://github.com/bazelbuild/examples/pull/563 solves this problem altogether.
Details in the PR description.

Please review when you find a minute.

Thank you

sgowroji:
We can close this issue as the ci is green now. https://buildkite.com/bazel/bazel-bazel-examples/builds/4216

PR #563, opened 2025-03-16
Fix 562, Rust example 07: vendoring

marvin-hansen:
This PR resolves issue https://github.com/bazelbuild/examples/issues/562

For once, the affected Build file has been updated to use the vendored Tokio crate alias without a version number
so in case the vendoring script ever downloads a different version the build should still succeed as long as the newer
crate version remains API compatible.

Then, the root cause has been eliminated by pinning the crate dependency to an exact version so that
Bazel should download exactly the pinned version.

The issue was caused by a very unlucky version convention in crate.spec in which

'version = "1.43.0",'

does not actually pin the version. Instead, it sets a lower version bound akin to "no lower than this version", 
which causes all sorts of problems. Instead, the correct form to pin a crate version is to use an equal sign prefix:

'version = "=1.43.0",

Complicating things further, the vendor script generates two aliases, one with a version and one without a version for each vendored crate.

As none of this is obvious let alone intuitive, I have updated the Readme to explain the intricate details
of the rules_rust vendoring behavior so that the reader can make an informed decision about vendoring, version pining,
and aliasing. 

Housekeeping:
Bumped up Bazel version to latest 8.1.1 for all Rust examples via the .bazelversion file.


meteorcloudy (approved):
Thanks!

meteorcloudy:
I guess we could revert https://github.com/bazelbuild/examples/pull/559

@sgowroji

PR #643, opened 2025-12-04
Fix build with Bazel last_green

fmeum:
Fixes: 
```
ERROR: /private/var/tmp/_bazel_fmeum/0465a0857e26a35c072f48ab751a1794/external/aspect_rules_jest+/jest/private/BUILD.bazel:11:1: name 'sh_binary' is not defined (did you mean 'cc_binary'?)
ERROR: /private/var/tmp/_bazel_fmeum/0465a0857e26a35c072f48ab751a1794/external/aspect_rules_jest+/jest/private/BUILD.bazel: no such target '@@aspect_rules_jest+//jest/private:jest_config_template.mjs': target 'jest_config_template.mjs' not declared in package 'jest/private' defined by /private/var/tmp/_bazel_fmeum/0465a0857e26a35c072f48ab751a1794/external/aspect_rules_jest+/jest/private/BUILD.bazel; however, a source file of this name exists.  (Perhaps add 'exports_files(["jest_config_template.mjs"])' to jest/private/BUILD.bazel?)
```

### DIFF
--- a/frontend/MODULE.bazel
+++ b/frontend/MODULE.bazel
@@ -2,7 +2,7 @@
 
 bazel_dep(name = "aspect_bazel_lib", version = "2.21.0")
 bazel_dep(name = "aspect_rules_lint", version = "1.10.2")
-bazel_dep(name = "aspect_rules_jest", version = "0.22.0")
+bazel_dep(name = "aspect_rules_jest", version = "0.24.2")
 bazel_dep(name = "aspect_rules_js", version = "2.0.2")
 bazel_dep(name = "aspect_rules_swc", version = "2.4.4")
 bazel_dep(name = "aspect_rules_ts", version = "3.1.0")


fmeum:
@gregestren Could you take a look?
@alexeagle FYI, I just found this while testing Bazel@HEAD, other such issues may remain.

fmeum:
Looks like CI is broken in other ways. CC @meteorcloudy

gregestren:
But as Im reading your comment, looks like there are new regressions since then...  

Since this PR updates to a newer version let's use this one.

gregestren:
> Looks like CI is broken in other ways. CC @meteorcloudy

I can replicate the `configurations` CI failure with the rolling release:

```sh
 $ USE_BAZEL_VERSION=rolling bazelisk build -- ... -//read_attr_in_transition:will-break -//cc_binary_selectable_copts:app_forgets_to_set_features -//cc_binary_selectable_copts:lib -//cc_binary_selectable_copts:app_forgets_to_set_features_native_binary -//cc_binary_selectable_copts:app_with_feature1_native_binary -//cc_binary_selectable_copts:app_with_feature2_native_binary -//auto_configured_builds/... -//cc_test/...
```

but it passes with ` USE_BAZEL_VERSION=last_green`. So I *think* we can just wait for the next rolling release and that part of CI will self-correct.

Same with `rules`.

So I think if we submit this PR, all CI will get green later this month.

alexeagle (approved):

Issue #644, opened 2025-12-08
[Bazel CI] ERROR: Bazel Analysis Error: no such target '@@glog+//:glog' referenced by '//:main'1


satyanandak:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/5127#019afc27-13ca-4209-bd7e-80f8f06acc18

Platform: Ubuntu, MacOS, Windows

Logs:

```
(04:11:13) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/57bc060e76715d43fc0b0818feed8755/external/glog+/BUILD.bazel: no such target '@@glog+//:glog': target 'glog' not declared in package '' defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/57bc060e76715d43fc0b0818feed8755/external/glog+/BUILD.bazel
(04:11:13) ERROR: /var/lib/buildkite-agent/builds/bk-docker-0svw/bazel-org-repo-root/examples/bzlmod/01-depend_on_bazel_module/BUILD:3:10: no such target '@@glog+//:glog': target 'glog' not declared in package '' defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/57bc060e76715d43fc0b0818feed8755/external/glog+/BUILD.bazel and referenced by '//:main'
(04:11:13) ERROR: Analysis of target '//:main' failed; build aborted: Analysis failed
(04:11:13) INFO: Elapsed time: 2.578s, Critical Path: 0.02s
(04:11:13) INFO: 1 process: 1 internal.
(04:11:13) ERROR: Build did NOT complete successfully
(04:11:13) FAILED:
```
Culprit:

Steps:
```
git clone https://github.com/bazelbuild/examples
git reset 1fc2622b5f1b41183c6118059ac57e4590d92de7 --hard
export USE_BAZEL_VERSION=9ba1edd5516721ffef3ad34e4f4c4e385e603320
bazel build --lockfile_mode=off //:main
```

CC Greenteam @kotlaja

PR #654, opened 2026-01-05
Bzlmod examples - Fix errors with glog for Bazel@HEAD

kotlaja:
Fixes https://github.com/bazelbuild/examples/issues/644.

`Glog` repo needed patches in multiple places since it's missing load rules statements for Bazel@HEAD. Still, since glog is archived (https://github.com/google/glog), I've just switched to logging from absl.


mcantimu:
> How does #649 relate?

I think `glog` library has been deprecated and we should use `ng-log` library according the deprecation notice in https://github.com/google/glog

meteorcloudy (approved):
Thanks!

Issue #657, opened 2026-01-08
[Bazel CI] ERROR: Analysis of //:catch2_test failed due to missing cc_library load in external dependency


deepalak56:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/5194#019b9bcb-dd1c-4a3c-984b-1bd3deac8648

Platform: MacOs, Ubuntu, Windows

Logs:

```
(04:10:45) ERROR: Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/7b1308b0cd04b4d9a7ee5d2a3eec1f16/external/catch2+/BUILD.bazel", line 65, column 11, in <toplevel>
		cc_library(
	File "/virtual_builtins_bzl/bazel/exports.bzl", line 40, column 9, in _removed_rule_failure
Error in fail:
         This rule has been removed from Bazel. Please add a `load()` statement for it.
         This can also be done automatically by running:
         buildifier --lint=fix <path-to-BUILD-or-bzl-file>
(04:10:45) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/7b1308b0cd04b4d9a7ee5d2a3eec1f16/external/catch2+/BUILD.bazel: no such target '@@catch2+//:catch2_main': target 'catch2_main' not declared in package '' defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/7b1308b0cd04b4d9a7ee5d2a3eec1f16/external/catch2+/BUILD.bazel
(04:10:45) ERROR: /var/lib/buildkite-agent/builds/bk-docker-z1qn/bazel-org-repo-root/examples/third-party-dependencies/BUILD:25:8: no such target '@@catch2+//:catch2_main': target 'catch2_main' not declared in package '' defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/7b1308b0cd04b4d9a7ee5d2a3eec1f16/external/catch2+/BUILD.bazel and referenced by '//:catch2_test'
Target //:buildifier up-to-date:
  bazel-bin/buildifier.bash
(04:10:45) ERROR: Analysis of target '//:catch2_test' failed; build aborted: Analysis failed
(04:10:45) INFO: Elapsed time: 2.589s, Critical Path: 0.03s
(04:10:45) INFO: 6 processes: 6 internal.
(04:10:45) ERROR: Build did NOT complete successfully
```

[...]

PR #658, opened 2026-01-08
Fix examples with Bazel@HEAD

kotlaja:
This PR fixes multiple errors from [the Downstream](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/5194#_):
1. Error with catch2 (mentioned in https://github.com/bazelbuild/examples/issues/657) - Fix is updating its version.
2. Error with loads in toolchains_llvm - Fix is updating its version.
3. Error with toolchains_protoc (`name 'proto_library' is not defined`) - Fix is removing `toolchains_protoc` since it was overriding an incompatible version of toolchain.
4. Error with `incompatible_use_toolchain_transition` - Seemed to be fixed with upgrade of toolchains_llvm.

Also, updated links for documentation which were not working.

### DIFF
--- a/.bazelci/tutorial-rust.yml
+++ b/.bazelci/tutorial-rust.yml
@@ -10,16 +10,12 @@ tasks:
     name: "Rust Cross Compilation"
     platform: ubuntu2204
     working_directory: ../rust-examples/02-hello-cross
-    build_flags:
-      - "--spawn_strategy=local"
     build_targets:
       - "//:all"
   rust-02-hello-cross-mac:
     name: "Rust Cross Compilation"
     platform: macos_arm64
     working_directory: ../rust-examples/02-hello-cross
-    build_flags:
-      - "--spawn_strategy=local"
     build_targets:
       - "//:all"
   rust-03-comp-opt-linux:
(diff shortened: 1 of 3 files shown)


gregestren:
> As an aside, does anyone know why the CI badge on https://github.com/bazelbuild/examples is red even though the [main branch](https://buildkite.com/bazel/bazel-bazel-examples) is consistently green?

I guess https://buildkite.com/bazel/bazel-bazel-examples/settings/badges explains it:

- https://badge.buildkite.com/260bbace6a4067a3c60539a31fed1191d341a24cb0bfeb0e23.svg   - red
- https://badge.buildkite.com/260bbace6a4067a3c60539a31fed1191d341a24cb0bfeb0e23.svg?branch=main - green

Which is the better signal for this repo?

kotlaja:
I didn't manage to reproduce the error but I figured out that `toolchains_llvm++llvm+llvm_toolchain/bin/clang-format` is a symlink that doesn't exist (pointing to nothing). The reason is inside LLVM 19.1.6 - This archive does not contain `bin/clang-format`, `bin/clang-tidy`, etc.  In other words, when Bazel tries to use or analyze this file (likely as an input to a sandbox), it sees a file type it "doesn't support" (a broken symlink) and fails. 

* Could be a "Why it fails with Bazel 9 but not with Bazel 8" question.. 

Bazel 9 potentially has stricter input validation for action execution?? Could be that some additional validation is added when it prepares the sandbox or inputs for compilation. Then it would encounter this broken symlink and throw an IOException because the file type (broken symlink) is invalid/unsupported in that context.

meteorcloudy (approved):
Thank you! \o/

PR #672, opened 2026-03-17
Add tsconfig paths example for next.js

navignaw:
Example to demonstrate the use of tsconfig paths.

* [rules_js docs](https://github.com/aspect-build/rules_js/blob/99ce5aa229a9900ca87e315e84614c40237a15cf/docs/faq.md?plain=1#L23-L26)
* [Example in rules_ts](https://github.com/aspect-build/rules_ts/blob/74d54bda208695d7e8992520e560166875cfbce7/examples/simple/tsconfig.json#L4-L10)
* [Next.js docs](https://nextjs.org/docs/14/app/building-your-application/configuring/absolute-imports-and-module-aliases)

### DIFF
--- a/frontend/next.js/tsconfig.json
+++ b/frontend/next.js/tsconfig.json
@@ -11,7 +11,11 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "baseUrl": ".",
+    "paths": {
+      "@components/*": ["components/*"],
+    }
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
   "exclude": ["node_modules"]
--- a/frontend/next.js/components/BUILD.bazel
+++ b/frontend/next.js/components/BUILD.bazel
@@ -1,0 +1,20 @@
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+
+SRCS = [
+    "button.tsx",
+]
+
+ts_project(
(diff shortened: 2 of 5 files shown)


navignaw:
Created this PR to investigate a potential bug in the tsconfig paths feature. Next.js doesn't seem to be able to find the `@components` import, even though the ts_project itself compiles fine.

https://buildkite.com/bazel/bazel-bazel-examples/builds/5211/steps/canvas?sid=019cfc36-892c-46a3-b33a-6b431c310e73&tab=output

frontend/next.js/tsconfig.json
```diff
@@ -11,7 +11,11 @@
...
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "baseUrl": ".",
+    "paths": {
+      "@components/*": ["components/*"],
+    }
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
```
jbedard, line 20: Does this mean nextjs won't apply the `paths` to the pre-compiled `.js` files?

jbedard:
I don't see the tsconfig file being passed to the `next()` invocation?
The nextjs config file should also reference the tsconfig like what was done in https://github.com/bazelbuild/examples/commit/f41da338ffc1736c86e6537966923c4adeef3854

navignaw:
couldn't get this working, but we ended up abandoning `tsconfig` paths and switching back to the fully qualified package name, which seems to work better and is already in the bazel examples.

PR #698, opened 2026-06-09
Add python-tutorial

molleer:
Adding a python tutorial which mirrors the cpp-tutorial.


molleer:
@gregestren @Wyverald could you have a look at this PR? It's been up for quite a while now but it has just been ignored..

PR #704, opened 2026-08-26
Remove the OCI parts of the java-maven example

ted-xie:
OCI is out of scope for an example that centers on Maven deps for Java builds. The OCI part should be moved to rules_oci.

See https://github.com/bazel-contrib/rules_oci/pull/905

Fixes a Bazel Downstream pipeline breakage.


ted-xie:
Still discussing this with Alex Eagle and co, so will convert to draft.

